Issue #53, opened 2019-10-31
Compass progressively increases in size


julianrex:
When `Bold Text` is enabled in `Settings -> Display & Brightness`, the compass view will progressively increase in size when the following code is repeatedly called:

```
mapView.compassViewMargins = margins
mapView.layoutSubviews()
```

m-stephen:
It seems that system will resize view's bounds when app set 'Bold Text'. Once the superview is calling `layoutSubviews`, the subview's bounds will progressively increase by this call stack:
```
    frame #0: 0x000000010291b330 Mapbox`::-[MGLCompassButton setBounds:](self=0x000000010e442420, _cmd="setBounds:", bounds=(origin = (x = 0, y = 0), size = (width = 41, height = 41))) at MGLCompassButton.mm:38:5
    frame #1: 0x00000001a6831098 UIKitCore`-[UIView(Geometry) _applyISEngineLayoutValuesToBoundsOnly:] + 532
    frame #2: 0x00000001a6831468 UIKitCore`-[UIView(Geometry) _resizeWithOldSuperviewSize:] + 140
    frame #3: 0x00000001a22857b0 CoreFoundation`__NSARRAY_IS_CALLING_OUT_TO_A_BLOCK__ + 24
    frame #4: 0x00000001a2183460 CoreFoundation`-[__NSArrayM enumerateObjectsWithOptions:usingBlock:] + 432
    frame #5: 0x00000001a6830568 UIKitCore`-[UIView(Geometry) resizeSubviewsWithOldSize:] + 172
    frame #6: 0x00000001a6774364 UIKitCore`-[UIView(AdditionalLayoutSupport) _is_layout] + 152
    frame #7: 0x00000001a6838374 UIKitCore`-[UIView(Hierarchy) _updateConstraintsAsNecessaryAndApplyLayoutFromEngine] + 1048
    frame #8: 0x000000010292a6c4 Mapbox`::-[MGLMapView layoutSubviews](self=0x000000010f003000, _cmd="layoutSubviews") at MGLMapView.mm:970:5
    frame #9: 0x000000010244071c MapExperience`-[ViewController viewWillAppear:](self=0x000000010e424860, _cmd="viewWillAppear:", animated=YES) at ViewController.m:129:5
    frame #10: 0x00000001a5d5bf48 UIKitCore`-[UIViewController _setViewAppearState:isAnimating:] + 652
[...]

PR #57, opened 2019-11-08
[ios] Fix ornaments' view constraints bug in bold-text mode

m-stephen:
close: #53

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -889,8 +889,11 @@ public:
     }
 
     if (!CGSizeEqualToSize(size, CGSizeZero)) {
-        [updatedConstraints addObject:[view.widthAnchor constraintEqualToConstant:size.width]];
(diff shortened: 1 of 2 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -939,10 +942,31 @@ - (void)installAttributionButtonConstraints {
...
-                                  size:self.attributionButton.bounds.size
+                                  size:[self sizeForOrnament:self.attributionButton constraints:self.attributionButtonConstraints]
                                margins:self.attributionButtonMargins];
 }
 
+- (CGSize)sizeForOrnament:(UIView *)view
+              constraints:(NSMutableArray *)constraints {
+    // avoid regenerating size constraints
+    CGSize size;
+    if(constraints && constraints.count > 0) {
```
fabian-guerra: This could be simplified to `if(constraints.count > 0) {`

Issue #52, opened 2019-10-31
Add scalebar width ratio and maximum width properties


julianrex:
The scale bar's maximum width is currently 1/2 of the view's width. We should

- Make this ratio configurable (a property?)
- Limit the width to some maximum (another configurable property). For example, with a default of 0.5 the scale bar on iPad could be considered too long.
- Decide how the scale bar's margins should affect this maximum width.

julianrex:
cc @chloekraw

m-stephen:
Some suggestions:
1. This ratio should have a minimum value. After testing, suggest 0.2 is a best one.
2. Don't suggest any limitation for maximum width for a specific value, it should be decided by users. Consider some developers may need this scale bar with full screen width at bottom of map view.
3. Margins should only affect the maximum width when the view's `max X` probably out of the screen.

@julianrex

tsuz:
@m-stephen 

In one scenario, if the ratio is set to 0.5 but then it is loaded in a SplitView where the width is 100px, in this case, the width of the scale bar should be close to 100px. If I set the maximum ratio to 1, would it update the width of the scale bar from 50px to 100px?

m-stephen:
Sorry for late reply.  @tsuz 

> In one scenario, if the ratio is set to 0.5 but then it is loaded in a SplitView where the width becomes 100px, in this case, the width of the scale bar should be close to 100px. If I set the maximum ratio to 1, would it dynamically update the width of the scale bar from 50px to 100px?

Correct. After this PR, the scale bar's width will be changed dynamically when you set the maximum ratio.

tsuz:
@m-stephen 

Thanks for confirming!

PR #65, opened 2019-11-20
Add a `maximumWidthRatio` property to customize maximum width of scale bar.

m-stephen:
This value is limited from 0.1 to 1.
Default value is 0.5

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [CONTRIBUTING.md](../../CONTRIBUTING.md) to get started.
 
+## master
+
+* Fix ornaments' view constraints bug when devices change to bold-text mode. ([#57](https://github.com/mapbox/mapbox-gl-native-ios/pull/57))
+
 ## 5.5.0
 
 ### Performance improvements
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -331,7 +331,7 @@ MGL_EXPORT
 
 /**
  The maximum width ratio of the scale bar relative to the map view's frame.
- This value is limited from 0.2 to 1.
+ This value is limited from 0.1 to 1.
  Default is 0.5(half of map view's width).
  */
 @property (nonatomic, assign) CGFloat scaleBarMaximumWidthRatio;
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -330,6 +330,13 @@ MGL_EXPORT
 @property (nonatomic, assign) CGPoint scaleBarMargins;
 
 /**
+ The maximum width ratio of the scale bar relative to the map view's frame.
+ This value is limited from 0.2 to 1.
+ Default is 0.5(half of map view's width).
+ */
+@property (nonatomic, assign) CGFloat scaleBarMaximumWidthRatio;
+
+/**
  A control indicating the map’s direction and allowing the user to manipulate
  the direction, positioned in the upper-right corner.
  */
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -576,6 +576,7 @@ public:
     _scaleBarConstraints = [NSMutableArray array];
     _scaleBarPosition = MGLOrnamentPositionTopLeft;
     _scaleBarMargins = MGLOrnamentDefaultPositionOffset;
+    _scaleBar.margins = MGLOrnamentDefaultPositionOffset;
 
     [self installConstraints];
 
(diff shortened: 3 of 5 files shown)


platform/ios/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 ## master
 
 * Fix ornaments' view constraints bug when devices change to bold-text mode. ([#57](https://github.com/mapbox/mapbox-gl-native-ios/pull/57))
+* Add a `maximumWidthRatio` property for customizing maximum width of scale bar. ([#65](https://github.com/mapbox/mapbox-gl-native-ios/pull/65))
```
fabian-guerra: What do you think if we rephrase this to:
> Added `MGLMapView.maximumWidthRatio` property for customizing the width ratio of scale bar. ([#65](https://github.com/mapbox/mapbox-gl-native-ios/pull/65))

platform/ios/src/MGLMapView.h
```diff
@@ -329,6 +329,13 @@ MGL_EXPORT
  */
 @property (nonatomic, assign) CGPoint scaleBarMargins;
 
+/**
+ The maximum width ratio of the scale bar relative to the map view's frame.
+ This value is limited from 0.1 to 1.
+ Default is 0.5(half of map view's width).
+ */
+@property (nonatomic, assign) CGFloat scaleBarMaximumWidthRatio;
```
fabian-guerra: What do you think if we rename this property to `scaleBarWidthRatio`?

platform/ios/src/MGLMapView.h
```diff
@@ -329,6 +329,13 @@ MGL_EXPORT
  */
 @property (nonatomic, assign) CGPoint scaleBarMargins;
 
+/**
+ The maximum width ratio of the scale bar relative to the map view's frame.
```
fabian-guerra: Seems that is not actually a maximum because you don't have an explicit range. e.g. minimum = 0.1 maximum = 1.0
The behavior for this property is to specify the width ratio.

platform/ios/src/MGLMapView.h
m-stephen: I think if changing this name, it might lead a wrong meaning. 
The value will only affect the maximum width of scale bar but could not equal to the actual width. 
The actual depends on the zoom level, and we can only say the scale bar's actual width is always less than or equal to the maximum width. 
@fabian-guerra

platform/ios/src/MGLMapView.h
m-stephen: As above I mentioned, this value is not for the actual width ratio, but only limiting the maximum width of scale bar(the actual width is dynamic and depends on zoom level)

platform/ios/src/MGLMapView.mm
```diff
@@ -800,9 +801,20 @@ - (void)setScaleBarPosition:(MGLOrnamentPosition)scaleBarPosition {
 - (void)setScaleBarMargins:(CGPoint)scaleBarMargins {
     MGLLogDebug(@"Setting scaleBarMargins: (x:%f, y:%f)", scaleBarMargins.x, scaleBarMargins.y);
     _scaleBarMargins = scaleBarMargins;
+    if([self.scaleBar isKindOfClass:[MGLScaleBar class]]) {
+        ((MGLScaleBar *)self.scaleBar).margins = scaleBarMargins;
+    }
     [self installScaleBarConstraints];
 }
 
+- (void)setScaleBarMaximumWidthRatio:(CGFloat)scaleBarMaximumWidthRatio {
```
tsuz: Can this test be included within `testChangeScaleBarSize`?

julianrex:
I'm not sure this PR solves the problem in https://github.com/mapbox/mapbox-gl-native-ios/issues/52 

e.g. 

- Set a width ratio of 0.3. 
- Run on iPad.
- In portrait mode, scale bar is acceptable
- In landscape, developer may consider it still to be too wide.

Does it?

What do you think about adding an optional *absolute* maximum width, measured in points?

Or are you suggesting that the developer modifies the ratio on rotation?

platform/ios/src/MGLMapView.mm
tsuz: Correct me if I'm wrong but I'm not sure if the tests cover the case where after this method is called, the scale bar adjusts accordingly.

m-stephen:
> Does it?
> 
> What do you think about adding an optional absolute maximum width, measured in points?
> 
> Or are you suggesting that the developer modifies the ratio on rotation?

Since we cannot know developers' UI requirements, I think developer should modify the ratio manually. @julianrex 

If developers display a map on a iPhone and support orientation, but one developer want a full screen width scale bar, but another want a half one. They both looks good in portrait mode, but when in landscape mode, if we make a limit absolute maximum width, it might be too short/long and out of expectations.

platform/ios/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [CONTRIBUTING.md](../../CONTRIBUTING.md) to get started.
 
+## Master
+
+* Add a `maximumWidthRatio` property for customizing maximum width of scale bar. ([#65](https://github.com/mapbox/mapbox-gl-native-ios/pull/65))
```
fabian-guerra: The blurb in changelogs should start with a verb in past tense. When you add a property you have to include the class name in order to link the documentation.
Please change the above to:
> Added \`MGLMapView.maximumWidthRatio\` property for customizing the width ratio of map view's scale bar.

platform/ios/src/MGLMapView.mm
```diff
@@ -800,9 +801,21 @@ - (void)setScaleBarPosition:(MGLOrnamentPosition)scaleBarPosition {
...
```
fabian-guerra: Please add the call to `MGLLogDebug`

chloekraw:
@m-stephen - is this ready for re-review?

habibahabiba77 (approved):

habibahabiba77 (commented):
Please let me know when I can get my back account and how I do it my maps open location

habibahabiba77 (approved):
[](https://api.mapbox.com/styles/v1/habibahabiba77/ckh2y671y258019p9p7u6otli.html?fresh=true&title=copy&access_token=pk.eyJ1IjoiaGFiaWJhaGFiaWJhNzciLCJhIjoiY2tnM3psd29lMGZycDMxcDRxMWZyY2x0dyJ9.18NwOEay84kyXrEkDZX9Jg)

lloydsheng:
@knov  @m-stephen has left. Does this function need to continue? I can take over the job.

/cc @julianrex

tsuz:
@m-stephen  @lloydsheng Long over due. Does the scalebar adjust accordingly when it's in a SplitView?

lloydsheng:
> Does the scalebar adjust accordingly when it's in a SplitView?

@tsuz  Per https://github.com/mapbox/mapbox-gl-native-ios/issues/52#issuecomment-565930806 and code, It should self-adjust.

tsuz (approved):
@lloydsheng  👍  Thank you!

PR #70, opened 2019-11-25
Move common configuration settings into shared xcconfig

julianrex:
This PR moves common settings into a shared `xcconfig` that includes an optional developer file.

My intention is that there should be an xcconfig for each of the configurations we have that includes this shared file.

We should also explore using the `developer.xcconfig` as a mechanism to add the Mapbox access token, without resorting to a custom build step.


nishant-karajgikar:
> We should also explore using the `developer.xcconfig` as a mechanism to add the Mapbox access token, without resorting to a custom build step.

This sounds neat! Would be nice to automate it all away with some sort of utility that logs into a users mapbox account, fetches the `access_token` and then populates the build settings in one fell swoop.

captainbarbosa:
>Would be nice to automate it all away with some sort of utility that logs into a users mapbox account, fetches the access_token and then populates the build settings in one fell swoop.

@nishant-karajgikar for some historical context there has been some discussion on how to best include a user's access token, but I believe we should highlight multiple ways for a user to do so (using a run script, setting in the `Info.plist` directly, or setting with `MGLAccountManager` etc.). With the above approach we would need to make two assumptions:

1) The only has one token they use with the app (what if they have multiple?)
2) What if they don't need to set a token at all? (not likely, but possible if they're not using any Mapbox data  in their map, since this is possible)

Setting a token can be a little weird for a first-time developer. I would instead propose a new Help doc outlining the different strategies one can use for setting an access token. But I wouldn't let that stop this PR!

nishant-karajgikar:
@captainbarbosa Ah okay, cool. Should've guessed that a bunch of thought has gone into these mechanisms! Considering those cases, good documentation probably is the best strategy. :) 

> 2. What if they don't need to set a token at all? (not likely, but possible if they're not using any Mapbox data  in their map, since this is possible)

Funnily, I know a couple of apps ([Citizen](https://apps.apple.com/us/app/citizen-protect-the-world/id1039889567), for one) that are doing this so it's not as uncommon as it may seem.

Issue #45, opened 2019-10-31
[iOS] Heading generate delay movement


matteocrippa:
Seems `didUpdate: heading` introduces delay in rendering from `didUpdate: location`.
Looking at code https://github.com/mapbox/mapbox-gl-native/blob/5c565a5e51be8379ddc2ffb5c859d8cfebe2fcd5/platform/ios/src/MGLMapView.mm#L5954 seems that once heading update has been called, this make a call for location too and this override the list of the location submit so far.

This is like kinda increasing the queue of the position and generate delayed movement.
The more you move the heading the more delays you get in movement of location.

### Steps to reproduce

 1. Activate heading and move with a location manager for position

### Expected behavior

Heading should react, but not generate delays on lat/lon.

### Actual behavior
Heading is like flooding the lat/lon and movements are delayed

### Configuration

**Mapbox SDK versions:**  5.4
**iOS/macOS versions:**  13.1
**Device/simulator models:** Simulator and Device
**Xcode version:** 11.1

matteocrippa:
@julianrex & @m-stephen make a PR that should mitigate and prob fix this #79

tomascern:
Hello, I have the same problem with the current version (6.2.1.). I am using followWithHeading tracking mode and custom annotation view (with rotation of image according to true heading) for user position and when the heading is changing, the user location is not updated, but when the heading is stable, the position jumps to the new location. Is there anything I can do to have simultaneous heading and position updates?

PR #79, opened 2019-11-29
Improve #45

matteocrippa:
I'm trying to mitigate issue #45, with this I tried moving the action to UI thread, I see some improvements but not total


matteocrippa:
@captainbarbosa I do an attempt to bump this, on my local implementation the performance has increased and XCode stop blocking me for background thread, think worth to consider this.
With stock one I was experiencing lot of lag and delay due to the issue #45

zugaldia:
Thank you @matteocrippa for the contribution and for the ping! There're a few other high-priority items that we're addressing but I'm adding this to our backlog to review.

PR #94, opened 2019-12-10
[ios] add allowScrollGesturesDuringRotateOrZoom

iBinh:
Add allowScrollGesturesDuringRotateOrZoom behavior like Google Map


platform/ios/src/MGLMapView.mm
```diff
@@ -1760,12 +1761,23 @@ - (void)handlePinchGesture:(UIPinchGestureRecognizer *)pinch
     [self cancelTransitions];
 
     CGPoint centerPoint = [self anchorPointForGesture:pinch];
+    if (!self.allowScrollGesturesDuringRotateOrZoom) {
+        if (pinch.numberOfTouches == 1 && pinch.state != UIGestureRecognizerStateEnded) {
```
jmkiley: Let's simplify this check. Since `centerPoint` is already equal to the pinch gesture's anchor point, `centerPoint` only needs to be changed if the conditions outlined aren't met.

iBinh:
@jmkiley I please review new commit

PR #101, opened 2019-12-13
Declare mbgl::style::GeoJSONOptions as immutable

captainbarbosa:
Attempts to fix crash in noted in https://github.com/mapbox/mapbox-gl-native/issues/16053 by declaring `mbgl::style::GeoJSONOptions` as an `Immutable` type.

Per chat with @asheemmamoowala, Android has implemented this [in their own platform JNI code](https://github.com/mapbox/mapbox-gl-native/blob/597757df6aba1ae8c4477706142823303f020de7/src/mbgl/style/sources/geojson_source_impl.cpp#L125) so they don't run into this issue. To be consistent, we should do the same on iOS.

So far, all of the changes in this PR seem to resolve the errors from https://github.com/mapbox/mapbox-gl-native/issues/16053, but I'm still running into an issue related to `lmbgl-vendor-icu` - running iosapp in this PR results in the build failing with `Library not found for -lmbgl-vendor-icu`:

Screenshot 2019-12-12 16 41 15

I'm not aware of what this library is used for. @asheemmamoowala has mentioned that @pozdnyakov may be able to diagnose this further.

Note: This PR also has the mapbox-gl-native PR pointed at the `release-tequila` branch.

cc @mapbox/ios


captainbarbosa:
Pulled over the finish line by https://github.com/mapbox/mapbox-gl-native-ios/pull/102

Issue #107, opened 2019-12-16
Add isHorizontalScrollingEnabled option to mapView


malwoodsantoro:
This is a feature request for the ability to easily **_disable horizontal scrolling_**. It’s currently possible to do this by 1) restricting the bounds of the map (defining a northeast coordinate with a latitude of 90 and a southwest coordinate with a latitude of -90), but the scrolling behavior is not smooth. With this approach it’s necessary to swipe straight up and down or you bump into the east and west bounds. 

It’s also possible to disable horizontal scrolling 2) using a custom `UIPanGestureRecognizer`. This proves challenging as it’s necessary to figure out how to gradually decrease the speed of the scroll to mimic the default scroll behavior. 

It would be great to expose a simple option on the mapView for this behavior

cc @captainbarbosa

Android equivalent: https://github.com/mapbox/mapbox-gl-native-android/issues/105

julianrex:
@captainbarbosa @malwoodsantoro can we mark this as closed?

malwoodsantoro:
👍 This has been implemented https://github.com/mapbox/mapbox-gl-native-ios/pull/108

PR #108, opened 2019-12-16
Introduce limited axis scrolling

captainbarbosa:
Addresses https://github.com/mapbox/mapbox-gl-native-ios/issues/107 by introducing a new option to limit vertical and horizontal scrolling.

Developers can limit scrolling for users both horizontally or vertically with this change.

Usage:

```objc
mapView.scrollingMode = MGLPanScrollingModeVertical
```

or 

```objc
mapView.scrollingMode = MGLPanScrollingModeHorizontal
```

Default:

```objc
mapView.scrollingMode = MGLPanScrollingModeDefault
```

Below is an example of setting `MGLPanScrollingModeVertical`:

scrollington_boat_factory

### DIFF
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -105,6 +105,15 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
     MGLUserTrackingModeFollowWithCourse,
 };
 
+typedef NS_ENUM(NSUInteger, MGLPanScrollingMode) {
+    /** The map allows the user to only scroll horizontally. */
+    MGLPanScrollingModeHorizontal               = 0,
+    /** The map allows the user to only scroll vertically. */
+    MGLPanScrollingModeVertical,
+    /** The map allows the user to scroll both horizontally and vertically. */
+    MGLPanScrollingModeDefault
+};
+
 /** Options for `MGLMapView.preferredFramesPerSecond`. */
 typedef NSInteger MGLMapViewPreferredFramesPerSecond NS_TYPED_EXTENSIBLE_ENUM;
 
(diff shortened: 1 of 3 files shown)


platform/ios/src/MGLMapView.h
```diff
@@ -105,6 +105,15 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
...
 
+typedef NS_ENUM(NSUInteger, MGLUserScrollingMode) {
+    /** The map allows the user to only scroll horizontally. */
+    MGLUserScrollingModeHorizontal              = 0,
+    /** The map allows the user to only scroll vertically. */
+    MGLScrollingModeVertical,
+    /** The map allows the user to scroll both horizontally and vertically. */
+    MGLScrollingModeUnlimited
+};
+
```
julianrex: Should this be `MGLUserScrollingMode` or `MGLScrollingMode`?

platform/ios/src/MGLMapView.mm
```diff
@@ -1721,7 +1722,18 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
...
+               case MGLScrollingModeVertical:
+                  self.mbglMap.moveBy({ 0, delta.y });
+                  break;
+               case MGLUserScrollingModeHorizontal:
+                  self.mbglMap.moveBy({ delta.x, 0 });
+                  break;
+               default:
+                  self.mbglMap.moveBy({ delta.x, delta.y });
+            }
+
```
julianrex: If you limited scrolling, then when the user scrolls in a perpendicular direction you still get some movement which feels a bit odd. 

What do you think about scaling the amount the map moves by, by a factor that takes this into account?

platform/ios/src/MGLMapView.h
julianrex: Given that this only affects the pan gesture, perhaps we include `Pan` in there somewhere?

Longer term, this property feels like it would better live with a custom pan gesture that we expose.

platform/ios/src/MGLMapView.mm
captainbarbosa: Per chat we'll defer this for a follow-up PR if needed.

captainbarbosa:
@julianrex I renamed some things, could you re-review please?

platform/ios/src/MGLMapView.mm
```diff
@@ -1726,11 +1727,23 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
     {
         CGPoint delta = [pan translationInView:pan.view];
 
+        NSLog(@"%@", NSStringFromCGPoint(delta));
```
julianrex: Let's remove this log.

Issue #48, opened 2019-10-31
Ensure ornaments are correctly positioned


julianrex:
Please patch https://github.com/mapbox/mapbox-gl-native/pull/14373 to this repo.

@m-stephen can you help here, if @lloydsheng doesn't have bandwidth?

PR #124, opened 2020-01-02
Add ornaments position asserts

lloydsheng:
The PR was a copy of https://github.com/mapbox/mapbox-gl-native/pull/14373, Fixed issue https://github.com/mapbox/mapbox-gl-native-ios/issues/48

/cc @julianrex @m-stephen

### DIFF
--- a/platform/ios/test/MGLMapViewLayoutTests.m
+++ b/platform/ios/test/MGLMapViewLayoutTests.m
@@ -5,6 +5,28 @@
 
 #import "MGLScaleBar.h"
 
+@interface UIView (MGLAdditions)
+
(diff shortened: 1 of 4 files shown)


platform/ios/test/MGLMapViewLayoutTests.m
```diff
@@ -5,6 +5,28 @@
...
 
+@interface UIView (MGLAdditions)
+
+- (CGRect)mgl_frameForIdentifyTransform;
+
+@end
+
+@implementation UIView (MGLAdditions)
+
+- (CGRect)mgl_frameForIdentifyTransform {
```
julianrex: Can we use the method from the SDK here, in case the two drift apart?

PR #130, opened 2020-01-03
Copy static headers from correct directory

bsudekum:
👋 Given the following command to create a static framework,

```
make iframework BUILD_DEVICE=false FORMAT=static BUILDTYPE=Release
```

headers will be outputed to the following path
```
build/ios/RelWithDebInfo-iphonesimulator/Headers
```

However when copying over the headers, we're looking in
```
build/ios/RelWithDebInfo-iphoneos/Headers
```

resulting in the following error:

```
cp: build/ios/RelWithDebInfo-iphoneos/Headers: No such file or directory
make: *** [iframework] Error 1
```

Source:
https://github.com/mapbox/mapbox-gl-native-ios/blob/fcc68b1c9057f74fcbf0e81666fc6b86800f6be0/platform/ios/scripts/package.sh#L239-L244

This PR makes it so depending on whether `BUILD_DEVICE` is true, we copy headers over from the appropriate directory.

### DIFF
--- a/platform/ios/scripts/package.sh
+++ b/platform/ios/scripts/package.sh
@@ -238,9 +238,9 @@ fi
 
 if [[ ${BUILD_STATIC} == true ]]; then
     step "Copying static library headers…"
-    cp -rv "${PRODUCTS}/${BUILDTYPE}-iphoneos/Headers" "${OUTPUT}/static/${NAME}.framework/Headers"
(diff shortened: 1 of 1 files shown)


julianrex:
Thanks @bsudekum!

julianrex:
Merging on behalf of @bsudekum.

PR #150, opened 2020-01-16
Fail the strip-frameworks script if required frameworks are not embedded.

julianrex:
This PR changes the `strip-frameworks.sh` script to fail if `Mapbox.framework` or `MapboxMobileEvents.framework` are not embedded (in the host application).

If this script fails, please make sure that the frameworks are selected as `Embed & Sign`

Embed Sign

### DIFF
--- a/platform/ios/framework/strip-frameworks.sh
+++ b/platform/ios/framework/strip-frameworks.sh
@@ -85,10 +85,12 @@ done
 
 if [[ -z $found_mapbox_sdk ]]; then
     echo "Please add Mapbox.framework to the Embed Frameworks build phase."
+    echo "See https://github.com/mapbox/mapbox-gl-native-ios/pull/150"
     exit -1
 fi
 
 if [[ -z $found_events_sdk ]]; then
     echo "Please add MapboxMobileEvents.framework to the Embed Frameworks build phase."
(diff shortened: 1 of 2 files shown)


julianrex:
cc @alfwatt

platform/ios/framework/strip-frameworks.sh
```diff
@@ -53,6 +53,18 @@ for file in $(find . -type f -perm +111); do
...
   fi
+
+  # Check for required Mapbox frameworks
+  framework=$(basename $file)
+
+  if [ "$framework" = "Mapbox" ]; then
+    found_mapbox_sdk=true
+  fi
+
+  if [ "$framework" = "MapboxMobileEvents" ]; then
```
1ec5: MapboxMobileEvents could be present but it could be an incompatible version. Perhaps check the framework bundle’s Info.plist for a version that matches the map SDK’s requirements (accounting for semver).

Incidentally, do we explicitly specify the supported or required MapboxNavigationNative versions anywhere in this repository? All I can find are references to “the latest version”, which could potentially become inaccurate.

julianrex:
Closing this experimental PR in favour of a podspec script & runtime check: #156

PR #159, opened 2020-02-10
Add property to show whether the snapshotter logo is displayed.

julianrex:
Adds a `BOOL` property to show whether the snapshotter logo is displayed.

Please note: 

> The Mapbox terms of service, which governs the use of Mapbox-hosted vector tiles and styles,
[requires](https://docs.mapbox.com/help/how-mapbox-works/attribution/) most Mapbox customers to display the Mapbox logo. If this applies to you, do not hide the logo or change its contents.

### DIFF
--- a/platform/darwin/src/MGLMapSnapshotter.h
+++ b/platform/darwin/src/MGLMapSnapshotter.h
@@ -121,13 +121,13 @@ MGL_EXPORT
 
 /**
  :nodoc:
- Whether the Mapbox logo is displayed.
+ Whether the Mapbox wordmark is displayed.
 
  @note The Mapbox terms of service, which governs the use of Mapbox-hosted
  vector tiles and styles,
  <a href="https://docs.mapbox.com/help/how-mapbox-works/attribution/">requires</a> most Mapbox
- customers to display the Mapbox logo. If this applies to you, do not
- hide the logo or change its contents.
+ customers to display the Mapbox wordmark. If this applies to you, do not
+ hide the wordmark or change its contents.
  */
 @property (nonatomic, readonly) BOOL showsLogo;
 
--- a/platform/darwin/src/MGLMapSnapshotter.h
+++ b/platform/darwin/src/MGLMapSnapshotter.h
@@ -120,6 +120,7 @@ MGL_EXPORT
 @property (nonatomic) CGFloat scale;
 
 /**
+ :nodoc:
  Whether the Mapbox logo is displayed.
 
  @note The Mapbox terms of service, which governs the use of Mapbox-hosted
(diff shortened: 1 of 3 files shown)


platform/darwin/src/MGLMapSnapshotter.h
```diff
@@ -119,6 +119,17 @@ MGL_EXPORT
  */
 @property (nonatomic) CGFloat scale;
 
+/**
```
fabian-guerra: This declaration is not necessary. As long as we have a private interface that declares the property it can be used.

fabian-guerra: If we decide to keep this declaration then lets remove the documentation.

platform/darwin/src/MGLMapSnapshotter.h
```diff
@@ -119,6 +119,18 @@ MGL_EXPORT
...
+ :nodoc:
+ Whether the Mapbox logo is displayed.
```
captainbarbosa: [Per our documentation](https://docs.mapbox.com/help/how-mapbox-works/attribution/), Mapbox attribution contains the Mapbox _wordmark_ and text _attribution_. To be consistent, you might want to consider changing this (and the property name) to "Mapbox wordmark".

julianrex: > It works, but stumbled getting it up and running with Swift, where I ended up needing to make an Objective-C class that overrides showsLogo to get it to work.

Yep, that's no good. Let me rethink.

julianrex:
> It works, but stumbled getting it up and running with Swift, where I ended up needing to make an Objective-C class that overrides showsLogo to get it to work.

Yep, that's no good. Let me rethink.

platform/darwin/src/MGLMapSnapshotter.h
julianrex: Done - I've kept the properties as `logo` for consistency with `MGLMapView` and Android.

platform/darwin/src/MGLMapSnapshotter.h
```diff
@@ -119,6 +119,18 @@ MGL_EXPORT
...
 
+/**
+ :nodoc:
+ Whether the Mapbox wordmark is displayed.
+
+ @note The Mapbox terms of service, which governs the use of Mapbox-hosted
+ vector tiles and styles,
+ <a href="https://docs.mapbox.com/help/how-mapbox-works/attribution/">requires</a> most Mapbox
+ customers to display the Mapbox wordmark. If this applies to you, do not
+ hide the wordmark or change its contents.
```
1ec5: Ignoring any internal options for a moment and focusing on this property declaration alone, it wouldn’t be possible for the developer to hide the wordmark or change its contents, so why admonish against doing so?

1ec5: It is always possible for the developer to obscure the logo by blotting it out or clipping the snapshot. But I think a warning against doing that should be somewhere more prominent, like in the class documentation.

PR #173, opened 2020-02-21
Update label languages for Mapbox Streets v8.3

### LINKED ISSUE #169: It is invalid to switch traditional languages
- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
    [style localizeLabelsIntoLocale:nil];
}

I find that other languages can be switched, so why can't traditional languages be switched.

1ec5:
Mapbox Streets source v8.3 adds names in Traditional Chinese and Vietnamese, according to [this documentation](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#common-fields). With this change, the `-[MGLStyle localizeLabelsIntoLocale:]` and `-[NSExpression mgl_expressionLocalizedIntoLocale:]` methods can now localize text into these languages (automatically by default if the system’s Preferred Language setting is set to one of them).

Fixes #169.

/cc @mapbox/maps-ios @lukasmartinelli @jcsg @lloydsheng

### DIFF
--- a/platform/darwin/src/MGLVectorTileSource.mm
+++ b/platform/darwin/src/MGLVectorTileSource.mm
@@ -89,8 +89,8 @@
  https://www.mapbox.com/vector-tiles/mapbox-streets-v8/
  */
 static NSArray * const MGLMapboxStreetsLanguages = @[
-    @"ar", @"de", @"en", @"es", @"fr", @"ja", @"ko", @"pt", @"ru", @"zh",
-    @"zh-Hans",
+    @"ar", @"de", @"en", @"es", @"fr", @"ja", @"ko", @"pt", @"ru", @"vi",
+    @"zh-Hans", @"zh-Hant",
 ];
 
 /**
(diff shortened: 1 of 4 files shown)


platform/darwin/src/MGLVectorTileSource.mm
```diff
@@ -89,17 +89,17 @@ @implementation MGLVectorTileSource (Private)
  https://www.mapbox.com/vector-tiles/mapbox-streets-v8/
  */
 static NSArray * const MGLMapboxStreetsLanguages = @[
-    @"ar", @"de", @"en", @"es", @"fr", @"ja", @"ko", @"pt", @"ru", @"zh",
-    @"zh-Hans",
+    @"ar", @"de", @"en", @"es", @"fr", @"ja", @"ko", @"pt", @"ru", @"vi",
```
1ec5: `-[NSExpression mgl_expressionLocalizedIntoLocale:]` has logic to fall back from `zh-Hant` to `zh` if the style happens to be using Streets source v7.

PR #178, opened 2020-02-25
Restore macOS map SDK; clean up iOS, macOS changelogs

1ec5:
This PR restores the macOS map SDK that had previously been maintained [in the Mapbox GL Native monorepo](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/macos/).

macosapp

The macOS build has over a hundred warnings due to a CMake misconfiguration in gl-native, which will be fixed by mapbox/mapbox-gl-native#16228. That PR doesn’t strictly block this one, however.

This PR also cleans up the changelogs for both iOS and macOS. The overgrown “master” section of the macOS changelog has been reorganized and copyedited. The iOS changelog now reflects the iOS map SDK v5.5.1 release and incorporates some of the same copyediting as the macOS changelog.

Fixes mapbox/mapbox-gl-native#13386.

/cc @mapbox/maps-ios

PR #179, opened 2020-02-25
Restore runtime styling code generation

1ec5:
Fixed the `darwin-style-code` make rule, which synchronizes the iOS and macOS map SDKs’ headers and implementation with the latest style specification.

mapbox/mapbox-gl-native#15177 suppressed three unimplemented properties from the style specification: `circle-sort-key`, `line-sort-key`, `fill-sort-key`. However, it did so in three different places, one per platform. mapbox/mapbox-gl-native#15839 and mapbox/mapbox-gl-native#15875 removed only the override that affects code generation in mbgl, but not the corresponding overrides for Android and iOS/macOS. mapbox/mapbox-gl-native#10937 created scripts/style-spec.js as a central place for these overrides, so this PR moves the overrides back there.

This PR also removes the overrides for `line-sort-key` and `fill-sort-key`, which were implemented in mapbox/mapbox-gl-native#15839. Now that the make rule works again, running it adds corresponding properties to the MGLFillStyleLayer and MGLLineStyleLayer classes.

Finally, the “Nitpick Darwin code generation” build step has been reenabled. It fails whenever the style specification changes in a way that would require us to rerun `make darwin-style-code`.

Depends on #178. Before merging, change the base from 1ec5-macos to master and rebase onto master.

/cc @mapbox/maps-ios @mapbox/gl-core


scripts/style-spec.js
```diff
@@ -1 +1,4 @@
-var spec = module.exports = require('../mapbox-gl-js/src/style-spec/reference/v8');
+var spec = module.exports = require('../vendor/mapbox-gl-native/mapbox-gl-js/src/style-spec/reference/v8');
+
+// FIXME: https://github.com/mapbox/mapbox-gl-native/issues/15008
+delete spec.layout_circle["circle-sort-key"];
```
1ec5: Ideally, gl-native would also centralize style specification overrides in scripts/style-spec.js so that this repository could simply require the file from that submodule instead of maintaining a separate copy.

nishant-karajgikar:
> Finally, the “Nitpick Darwin code generation” build step has been reenabled. It fails whenever the style specification changes in a way that would require us to rerun make darwin-style-code.

Awesome.

PR #181, opened 2020-02-27
Mapbox GL Native v1.3.0

1ec5:
Upgraded Mapbox GL Native from release-unicorn to [v1.3.0](https://github.com/mapbox/mapbox-gl-native/releases/tag/maps-v1.3.0):

 * Added the `in` expression function for testing whether a value is included in an array expression or whether a string is a substring of another string. Use this function in expressions in style JSON or with the `MGL_FUNCTION()` syntax in an `NSExpression` format string. (mapbox/mapbox-gl-native#16162)
* Added the `within` expression function for testing whether the evaluated feature lies within the given GeoJSON object. Use this function in expressions in style JSON or with the `MGL_FUNCTION()` syntax in an `NSExpression` format string. (mapbox/mapbox-gl-native#16157)
* Added the `MGLLineStyleLayer.lineSortKey` and `MGLFillStyleLayer.fillSortKey` properties. (#179, mapbox/mapbox-gl-native#16194, mapbox/mapbox-gl-native#16220)
* Icons and labels near tile boundaries draw more consistently. (mapbox/mapbox-gl-native#16159)
* Fixed an issue where an `MGLSymbolStyleLayer.lineDashPattern` value of `{1, 0}` resulted in hairline gaps. (mapbox/mapbox-gl-native#16202)
* Fixed issues where an offline pack would stop downloading before completion. (mapbox/mapbox-gl-native#16230, mapbox/mapbox-gl-native#16240)
* When an offline pack encounters an HTTP 404 error, the `MGLOfflinePackUserInfoKeyError` user info key of the `MGLOfflinePackErrorNotification` now indicates the resource that could not be downloaded. (mapbox/mapbox-gl-native#16240)
* Fixed a memory leak when zooming with any options enabled in the `MGLMapView.debugMask` property. (mapbox/mapbox-gl-native#15179)

Depends on mapbox/mapbox-gl-native#16242.

/cc @mapbox/maps-ios @mapbox/gl-native

PR #182, opened 2020-02-27
Nine-part resizable images

1ec5:
The `MGLSymbolStyleLayer.iconTextFit` property now respects the cap insets of any nine-part stretchable image passed into the `-[MGLStyle setImage:forName:]` method.

The “Manipulate Style” command in macosapp demonstrates the stretchable image feature by adding a single [Ohio state route shield image](https://commons.wikimedia.org/wiki/File:OH-blank.svg) (from a PDF with cap insets in the asset catalog) and labeling each Ohio state route with that image, stretched to fit the route number without widening the icon’s stroke.

ohio

Depends on #181. Retarget and rebase this PR onto master before merging.

/ref mapbox/mapbox-gl-native#16030
/cc @mapbox/maps-ios @kkaefer


1ec5:
I’ve added unit tests. The NSImage and UIImage extensions now round-trip cap insets (the inverse of the content rect). Note that NSImage and UIImage only support a single stretchable region, not the multiple stretchable regions introduced in mapbox/mapbox-gl-native#16030 for sprite sheets.

platform/ios/src/UIImage+MGLAdditions.mm
```diff
@@ -39,10 +53,30 @@ - (nullable instancetype)initWithMGLPremultipliedImage:(const mbgl::Premultiplie
...
+        self.capInsets.left / self.scale, (self.size.width - self.capInsets.right) / self.scale,
+    }};
+    mbgl::style::ImageStretches stretchY = {{
+        self.capInsets.top / self.scale, (self.size.height - self.capInsets.bottom) / self.scale,
+    }};
+    
+    mbgl::optional<mbgl::style::ImageContent> imageContent;
+    if (!MGLEdgeInsetsIsZero(self.capInsets)) {
+        imageContent = (mbgl::style::ImageContent){
+            .left = static_cast<float>(self.capInsets.left * self.scale),
```
julianrex: I'm confused by the `* self.scale` here, compared with the `/ self.scale` above. What's going on here? (The macOS version below has `*` in both)

1ec5: `-initWithMGLStyleImage:` divides by the scale and `-mgl_styleImageWithIdentifier:` multiplies by the scale on both platforms.

1ec5: Oh, I see what you mean – you’re referring to the `stretchX` and `stretchY` variables. Yes, I think the division above is incorrect. The tests aren’t catching the mistake because they only assert on the image’s content rect, not its stretchable areas per se.

1ec5:
Note that NSImage and UIImage only support a single resizable region (though it can be tiled as an alternative to stretching). There are third-party solutions for native support for multiple stretchable regions, but the usual way to implement such images is to draw the image programmatically on demand.

PR #183, opened 2020-02-27
Remove “in” expression workaround

1ec5:
Removed the `match`-based workaround for `in` expressions in favor of real support for an `in` expression operator, including support for testing whether a string is a substring of another string.

* [x] Replace `match`-based workaround with `in` operator mappings
* [x] Test substring matching
* [x] Implement `ANY`/`SOME` and `NONE` with `=` and `ALL` with `!=`
* [x] Document `in` expressions in “Information for Style Authors” guide
* [x] Document comparison modifiers in “Predicates and Expressions” guide

Fixes #168. Depends on #181. Retarget and rebase this PR onto master before merging.

/cc @mapbox/maps-ios

PR #184, opened 2020-02-29
Allow predicates to test whether a feature lies within a given shape

### LINKED ISSUE #175: Platform bindings for `within` expression
Tracking the development of platform bindings for the new `within` expression (https://github.com/mapbox/mapbox-gl-native/pull/16157; https://github.com/mapbox/mapbox-gl-native/pull/16194)

If it's possible, because this isn't too complex, it would be great to get this merged this week before the next `alpha` release.

cc @julianrex @zmiao

1ec5:
Added support for the NSPredicate syntax `SELF IN shape` or `shape CONTAINS SELF`, where `shape` is an MGLShape object or MGLFeature instance. These syntaxes are interpreted as `within` expressions in JSON format.

Fixed an issue where GeoJSON geometries were converted to MGLFeature instances instead of MGLShape objects. This issue probably hasn’t come up in the wild because the conversion code was previously only part of a feature querying code path, and feature querying methods return features anyways. But since it’s possible for the `shape` in `SELF IN shape` to be an MGLShape, that shape should get round-tripped as an MGLShape rather than an MGLFeature.

Fixes #175. Depends on #183, which in turn depends on #181. Retarget and rebase this PR onto master before merging.

/cc @mapbox/maps-ios


platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -659,11 +668,24 @@ + (instancetype)expressionWithMGLJSONObject:(id)object {
...
         return [NSExpression expressionForConstantValue:object];
     }
     
     if ([object isKindOfClass:[NSDictionary class]]) {
+        if (object[@"type"]) {
+            NSError *error;
+            NSData *shapeData = [NSJSONSerialization dataWithJSONObject:object options:0 error:&error];
+            MGLShape *shape;
+            if (shapeData && !error) {
+                shape = [MGLShape shapeWithData:shapeData encoding:NSUTF8StringEncoding error:&error];
```
1ec5: This works, but there may be a more efficient way to convert this JSON-like NSDictionary structure into an MGLShape. Perhaps we could first convert the NSDictionary structure into a C++ representation of the GeoJSON object, then convert to an MGLShape.

chloekraw:
@1ec5 will there be a developer-facing API that references `within` in its name?

1ec5:
> will there be a developer-facing API that references `within` in its name?

The `SELF IN` syntax is available for developers to use like this (notice lovely highlighting of NSExpression format string syntax):

```objc
[NSPredicate predicateWithFormat:@"SELF IN %@", shape];
```

But if you’re referring to something more reminiscent of the JSON expression format, it’s possible to use `within` expressions directly without this PR:

```objc
[NSPredicate predicateWithFormat:@"MGL_FUNCTION('within', %@)", shape];
```

or:

```objc
[NSPredicate predicateWithMGLJSONObject:@[@"within", @{@"type": @"Polygon", @"coordinates": @[@[@[@0, @0], @[@0, @1], @[@1, @1], @[@1, @0]]]}]];
```

I suppose we could install an aftermarket function like `mgl_within:` to avoid `MGL_FUNCTION()`, but it I’m not sure of the use case given the `IN` and `CONTAINS` operators.

platform/ios/docs/guides/For Style Authors.md
```diff
@@ -354,6 +354,7 @@ In style specification | Method, function, or predicate type | Format string syn
 `case`                 | `+[NSExpression expressionForConditional:trueExpression:falseExpression:]` or `MGL_IF` or `+[NSExpression mgl_expressionForConditional:trueExpression:falseExpresssion:]` | `TERNARY(1 = 2, YES, NO)` or `MGL_IF(1 = 2, YES, 2 = 2, YES, NO)`
 `coalesce`             | `mgl_coalesce:` | `mgl_coalesce({x, y, z})`
 `match`                | `MGL_MATCH` or `+[NSExpression mgl_expressionForMatchingExpression:inDictionary:defaultExpression:]` | `MGL_MATCH(x, 0, 'zero match', 1, 'one match', 'two match', 'default')`
+`within`               | `NSInPredicateOperatorType` | `SELF IN %@` or `%@ CONTAINS SELF` where `%@` is an `MGLShape`
```
1ec5: As always, if a developer is already familiar with the style specification’s JSON expression format or a similar syntax on Android, they can consult the “Information for Style Authors” guide to learn the corresponding NSExpression syntax.

1ec5:
Per https://github.com/mapbox/mapbox-gl-native-ios/issues/175#issuecomment-593371175, MGLConversion needs to convert from `mbgl::GeoJSON` to MGLShape:

https://github.com/mapbox/mapbox-gl-native-ios/blob/534d64c0a8e5084e0da8780530bbb78f25a0ea22/platform/darwin/src/MGLConversion.h#L127-L130

This code is called as part of `-[NSPredicate(MGLAdditions) mgl_filter]`, the lack of a conversion probably explains why I was unable to preserve the filter roundtripping assertions as part of the unit tests in this PR.

platform/darwin/src/NSExpression+MGLAdditions.mm
fabian-guerra: Is there any performance considerations for bridging into an C++ object then back to `MGLShape`? Is this something that is going to be called frequently?

1ec5: In the current implementation, `MGLStyleValueTransformer::PropertyExpressionEvaluator` converts the C++ `mapbox::geojson` struct to a GeoJSON-formatted NSDictionary, then this method converts it to a string, then back to a C++ `mapbox::geojson` struct, then finally to an MGLShape object. It is more roundabout than it should be.

This conversion process happens once when setting a layer’s predicate or paint/layout property (which can happen automatically e.g. when localizing the style). mbgl will evaluate the already-converted expression on each feature, but no SDK-side code will run as part of the rendering loop.

Ideally, mbgl would preserve the `mapbox::geojson` struct so that `MGLJSONObjectFromMBGLValue()` could convert it directly to an MGLShape. Absent such a change in mbgl, there are two alternative, purely SDK-side approaches:

* Convert directly to C++ `mapbox::geojson` structs then to MGLShape, skipping NSJSONSerialization
* Directly initialize MGLShapes using new initializers that parse the NSDictionary, skipping C++

but I’m inclined to treat either optimization as tail work for when we know more about how this expression operator will be used. The performance characteristics could differ depending on the complexity of the passed-in GeoJSON feature.

1ec5:
As of 11424ded73745e734c2d1f58b9fad8d7bd8ae635, the Debug ‣ Manipulate Style command adds a circle layer that should mark each POI [around here](https://www.openstreetmap.org/#map=16/39.1314/-84.5158). However, nothing shows up because the `within` expression always evaluates to `false`. The conversion from NSExpression to `mbgl::style::expression::Within` is working correctly, as is the conversion from GeoJSON-formatted NSDictionary to `mapbox::geometry`. Here’s what the GeoJSON representation of the shape now that `toGeoJSON()` has been implemented:

```
{"type":"Polygon","coordinates":[[[-84.520899999999997,39.127899999999997],[-84.511200000000002,39.127299999999998],[-84.510199999999998,39.1355],[-84.521199999999993,39.136000000000003],[-84.520899999999997,39.127899999999997]]]}
```

geojson.io

`Within::evaluate()` returns false because `pointsWithinPolygons()` returns false because `boxWithinBox()` returns false because `isBBoxValid()` returns false because `Within.polygonBBox` is never set. Something in the call to `convertGeometry()` causes the polygon’s only ring to go missing. In [this part of the method](https://github.com/mapbox/mapbox-gl-native/blob/90461930e4e7b8ce9901f9c41be47de8ab40d3f5/src/mbgl/tile/geometry_tile_data.cpp#L154-L164), `pg` is empty.

As far as I can tell, the SDK code is behaving correctly and the shape is valid, but the expression never evaluates to true. I’ve also tried reversing the polygon and removing the closing point to no avail.

/cc @zmiao

platform/darwin/docs/guides/Predicates and Expressions.md
```diff
@@ -55,6 +55,18 @@ The following aggregate operators are supported:
 `NSInPredicateOperatorType`       | `key IN { 'iOS', 'macOS', 'tvOS', 'watchOS' }`
 `NSContainsPredicateOperatorType` | `{ 'iOS', 'macOS', 'tvOS', 'watchOS' } CONTAINS key`
 
+You can use the `IN` and `CONTAINS` operators to test whether a value appears in a collection, whether a string is a substring of a larger string, or whether the evaluated feature (`SELF`) lies within a given `MGLShape` or `MGLFeature`. For example, to show one delicious local chain of sandwich shops, but not similarly named steakhouses and pizzerias:
```
1ec5: MGLFeature support was added in mapbox/mapbox-gl-native#16232, which we haven’t pulled into this repository yet.

1ec5:
> However, nothing shows up because the `within` expression always evaluates to `false`.

This will also be fixed by mapbox/mapbox-gl-native#16232 once we pull it in.

1ec5:
>> However, nothing shows up because the `within` expression always evaluates to `false`.
> 
> I’ve confirmed that this will also be fixed by mapbox/mapbox-gl-native#16232 once we pull it in.

#210 will pull in the fix. No additional changes are needed for `within` support at the SDK level, but this PR depends on #183, which is waiting for #210.

PR #188, opened 2020-02-29
Add async completion handler to MGLOfflineStorage::preloadData

alexshalamov:
As the database operations were moved to separate thread to avoid blocking resource loading, it might be required to know when preloadData has been completed.

### DIFF
--- a/platform/darwin/src/MGLOfflineStorage.h
+++ b/platform/darwin/src/MGLOfflineStorage.h
@@ -143,6 +143,15 @@ typedef void (^MGLOfflinePackRemovalCompletionHandler)(NSError * _Nullable error
 typedef void (^MGLBatchedOfflinePackAdditionCompletionHandler)(NSURL *fileURL, NSArray<MGLOfflinePack *> * _Nullable packs, NSError * _Nullable error);
 
 /**
+A block to be called once the data  has been preloaded.
+
+ @param url The URL of the data that was pre-loaded.
+ @param error Contains a pointer to an error object (if any) indicating why the
+ data could not be pre-loaded.
+*/
+typedef void (^MGLOfflinePreloadDataCompletionHandler)(NSURL * url, NSError * _Nullable error);
+
+/**
  The type of resource that is requested.
  */
 typedef NS_ENUM(NSUInteger, MGLResourceKind) {
@@ -434,7 +443,9 @@ MGL_EXPORT
  This method is asynchronous; the data may not be immediately available for
  in-progress requests, though subsequent requests should have access to the
  cached data.
- 
+
+ To find out when the resource is ready to retrieve from the cache, use the -preloadData:forURL:modificationDate:expirationDate:eTag:mustRevalidate:completionHandler: method.
+
  @param data Response data to store for this resource. The data is expected to
     be uncompressed; internally, the cache will compress data as necessary.
  @param url The URL at which the data can normally be found.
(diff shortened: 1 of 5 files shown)


alexshalamov:
@1ec5 do you think completion handler needs to be exposed publicly or we can have it as an internal method?

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -448,6 +448,37 @@ MGL_EXPORT
 
 - (void)putResourceWithUrl:(NSURL *)url data:(NSData *)data modified:(nullable NSDate *)modified expires:(nullable NSDate *)expires etag:(nullable NSString *)etag mustRevalidate:(BOOL)mustRevalidate __attribute__((deprecated("", "-preloadData:forURL:modificationDate:expirationDate:eTag:mustRevalidate:")));
 
+
+/**
+A block to be called once the data  has been preloaded.
+*/
+typedef void (^MGLOfflinePreloadDataCompletionHandler)(void);
```
1ec5: Where possible, a completion handler should accept some contextual information as parameters so that the developer can reuse their completion block code for multiple calls to the method. In this case, it would probably be helpful to pass in the URL and maybe the data.

1ec5: Do you think there’s a potential for the preloading operation to fail with an error, either now or in a future change? If so, let’s add an NSError-typed error parameter here to avoid a backwards-incompatible change.

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -448,6 +448,37 @@ MGL_EXPORT
...
+/**
+Inserts the provided resource into the ambient cache.
+
+This method mimics the caching that would take place if the equivalent resource
+were requested in the process of map rendering. Use this method to pre-warm the
+cache with resources you know will be requested.
+
+This method is asynchronous; the data may not be immediately available for
+in-progress requests, though subsequent requests should have access to the
+cached data.
```
1ec5:
```suggestion
This method is asynchronous. The data is available for in-progress requests as
soon as the completion handler is called.
```

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -448,6 +448,37 @@ MGL_EXPORT
...
 - (void)putResourceWithUrl:(NSURL *)url data:(NSData *)data modified:(nullable NSDate *)modified expires:(nullable NSDate *)expires etag:(nullable NSString *)etag mustRevalidate:(BOOL)mustRevalidate __attribute__((deprecated("", "-preloadData:forURL:modificationDate:expirationDate:eTag:mustRevalidate:")));
 
+
+/**
+A block to be called once the data  has been preloaded.
+*/
+typedef void (^MGLOfflinePreloadDataCompletionHandler)(void);
+
+/**
+Inserts the provided resource into the ambient cache.
```
1ec5: Distinguish this method from the original one:

```suggestion
Inserts the provided resource into the ambient cache, calling a completion handler when finished.
```

And add a cross-reference like this to the original method’s documentation comment, to the paragraph about asynchronicity:

> To find out when the resource is ready to retrieve from the cache, use the `-preloadData:forURL:modificationDate:expirationDate:eTag:mustRevalidate:completionHandler:` method.

platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -649,8 +658,14 @@ - (void)preloadData:(NSData *)data forURL:(NSURL *)url modificationDate:(nullabl
...
     }
-    
-    _mbglDatabaseFileSource->put(resource, response);
+
+    if (completion) {
+        _mbglDatabaseFileSource->forward(resource, response, [completion] {
+            dispatch_async(dispatch_get_main_queue(), completion);
+        });
+    } else {
+        _mbglDatabaseFileSource->put(resource, response);
```
1ec5: Is `put()` still asynchronous? What’s the difference between calling `forward()` with a no-op lambda and calling `put()`?

1ec5:
> do you think completion handler needs to be exposed publicly or we can have it as an internal method?

In general, the public API should accept a completion handler or delegate or send a notification for any asynchronous operation if possible.

platform/darwin/src/MGLOfflineStorage.mm
alexshalamov: > What’s the difference between calling forward() with a no-op lambda and calling put()

`put` is still asynchronous. `forward` with callback was introduced to keep compatibility for an old `put` method.

platform/darwin/src/MGLOfflineStorage.h
alexshalamov: done

platform/darwin/src/MGLOfflineStorage.h
alexshalamov: done

platform/darwin/src/MGLOfflineStorage.h
alexshalamov: done. Completion handler has url and an error.

alexshalamov: done

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -448,6 +450,40 @@ MGL_EXPORT
...
 
+
+/**
+A block to be called once the data  has been preloaded.
+
+ @param url The URL of the data that was pre-loaded.
+ @param error Contains a pointer to an error object (if any) indicating why the
+ data could not be pre-loaded.
+*/
+typedef void (^MGLOfflinePreloadDataCompletionHandler)(NSURL * url, NSError * _Nullable error);
```
1ec5: Nit: It’s more common for typedefs to be declared at the top of the file outside the interface declaration. Declaring it here doesn’t break Clang or jazzy but could be harder to find when looking for typedefs in the usual place.

platform/darwin/src/MGLOfflineStorage.mm
1ec5: Since `put` doesn’t have any advantage over `forward`, should we mark the preexisting public method as deprecated, to make it easier to remove in the future?

1ec5:
Also a question about deprecation: https://github.com/mapbox/mapbox-gl-native-ios/pull/188#discussion_r390432714.

platform/darwin/src/MGLOfflineStorage.mm
alexshalamov: @1ec5 sure, we can plan deprecation on native side.

alexshalamov:
@1ec5 @julianrex moved completion handler definition. Please free to merge if needed.

PR #204, opened 2020-03-10
Notarize macosapp

1ec5:
Notarize macosapp when packaging a release of the macOS map SDK, so macOS users can more easily open the application without jumping through Gatekeeper hoops.

To do:

* [x] Opt into notarization when using ExportOptions.plist
* [x] Add the Hardened Runtime entitlement to the project
* [x] Set `ENABLE_HARDENED_RUNTIME=YES` in invocations of `xcodebuild` in package.sh (as opposed to inside the project, so non–Mapbox team members can continue to build ad-hoc)
* [ ] [Add a post-archive action](https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution/customizing_the_notarization_workflow?language=objc) to the scheme that exports the application for notarization
* [ ] Invoke `xcrun altool` in package.sh
* [ ] Split out a separate `make` rule that invokes `xcrun stapler staple` and completes the deployment process, to be run after notarization succeeds

Fixes mapbox/mapbox-gl-native#14753. Reprise of mapbox/mapbox-gl-native#14754.

/cc @mapbox/maps-ios @frederoni

### DIFF
--- a/platform/macos/ExportOptions.plist
+++ b/platform/macos/ExportOptions.plist
@@ -2,8 +2,12 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>destination</key>
+	<string>upload</string>
 	<key>method</key>
 	<string>developer-id</string>
+	<key>signingStyle</key>
+	<string>automatic</string>
 	<key>teamID</key>
 	<string>GJZR2MEM28</string>
 </dict>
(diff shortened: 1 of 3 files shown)


1ec5:
> Set `ENABLE_HARDENED_RUNTIME=YES` in invocations of `xcodebuild` in package.sh (as opposed to inside the project, so non–Mapbox team members can continue to build ad-hoc)

I ended up moving this setting to the project. It seems to be possible to build the application ad-hoc with the hardened runtime setting after all.

1ec5:
> I ended up moving this setting to the project. It seems to be possible to build the application ad-hoc with the hardened runtime setting after all.

But then macosapp won’t run in the debugger. So I limited the hardened runtime setting to the Release configuration for now, but I’ll probably go back to setting the build setting in a script once everything is fully automated.

PR #210, opened 2020-03-13
Streamline MGLMapSnapshotter; upgrade to Mapbox GL Native v1.4.0

1ec5:
Upgraded MGLMapSnapshotter to be compatible with [mbgl v1.4.0](https://github.com/mapbox/mapbox-gl-native/releases/tag/maps-v1.4.0). For the iOS map SDK v5.8.0-beta.1 and macOS map SDK v0.15.0-beta.1 release notes, we’ll add:

* `MGLMapSnapshotter` no longer keeps itself from being deallocated before its completion handler is called. To ensure that the completion handler passed into `-[MGLMapSnapshotter startWithCompletionHandler:]` is called, maintain a strong reference to the snapshotter from a longer-lived class, such as the class where you call `-[MGLMapSnapshotter startWithCompletionHandler:]`. (#210)
* The `-[MGLMapSnapshotter cancel]` method no longer calls the completion handler passed into `-[MGLMapSnapshotter startWithCompletionHandler:]`. (#210)
* Fixed an issue where the `MGLMapSnapshotter.loading` property always returned `NO`, even while loading a snapshot. (#210)
* Downloaded offline packs no longer reduce the storage space available for ambient caching of tiles and other resources. (mapbox/mapbox-gl-native#15622)
* Improved performance when continuously animating a tilted map. (mapbox/mapbox-gl-native#16287)

Converted as many of MGLMapSnapshotter’s instance variables and properties as possible into local variables close to the point of use. `-loading` once again works and is used to avoid concurrent requests, now that the completion handler is no longer memoized. Snapshotter options are converted into an `mbgl::MapSnapshotter` on demand as late as possible.

The completion handler of `-startWithCompletionHandler:` and its variants is no longer called as part of `-cancel`, when the snapshotter is deallocated, or when the application is terminating, per the rationale in https://github.com/mapbox/mapbox-gl-native-ios/issues/200#issuecomment-598458116. To ensure that the snapshotter runs to completion, it is necessary to hold a strong reference to the snapshotter, ideally in the containing class.

Pushed some drawing and completion code up from inner completion blocks to outer completion blocks to avoid excessively passing state around. Converted a few class methods into standalone C functions to emphasize the avoidance of captured state.

More to come:

* [x] Upgrade mbgl to master
* [x] Regenerate runtime styling documentation
* [x] Update snapshotter to build against latest mbgl
* [x] Eliminate most snapshotter ivars
* [x] Simplify callbacks
[...]

Issue #200, opened 2020-03-07
Platform bindings for runtime styling APIs on `MapSnapshotter`


chloekraw:
https://github.com/mapbox/mapbox-gl-native/pull/16268

1ec5:
mapbox/mapbox-gl-native#16268 has yet to be released. In the meantime, I’ll update the submodule to master locally to work on the integration.

1ec5:
Changes to mbgl’s public API in mapbox/mapbox-gl-native#16268 are forcing me to rewrite MGLMapSnapshotter. It’s a long-overdue rewrite. I think we’ll be able to clear away many of the pessimizations and unsafe thread safety patches we’ve applied to this class over the years by removing indirection and parallel state.

The refactoring centers around passing the completion handler around in local scope without storing it as an ivar. It has been going well, except for one detail: if you pass a completion handler into `-startWithCompletionHandler:`, then call `-cancel`, the completion handler gets called from a class method. I would like to remove this behavior so that `-cancel` simply stops all activity. Apple’s asynchronous, NSOperation-friendly APIs similarly avoid calling the completion handler when canceling; examples include MKMapSnapshotter and MKDirections. If we don’t call the completion handler when canceling, then there’s no need to store the completion handler anywhere in SDK code, avoiding a whole class of memory and concurrency issues.

As far as I can tell, we don’t document anywhere that the completion handler does get called in this situation, so it would be a backwards-compatible change. The developer has full control over what happens when they call `-cancel`; they’re free to call that completion block themselves.

1ec5:
For the actual runtime styling feature, I plan to add an “actions” or “style handler” block to either `-[MGLMapSnapshotter startWithQueue:overlayHandler:completionHandler:]` or MGLMapSnapshotOptions that accepts an MGLStyle. This would afford more flexibility than mapbox/mapbox-gl-native#16286, which only allows the developer to add a single layer, and would be consistent with familiar APIs like `+[UIGraphicsImageRenderer imageWithActions:]` and `+[NSImage imageWithSize:flipped:drawingHandler:]`. In this case, a block is more appropriate than a full-on delegate protocol, because we only expect the style to load a single time per snapshot.

If we need to support the `onDidFailLoadingStyle` and `onStyleImageMissing` events, then we 
[...]

PR #211, opened 2020-03-14
Keep MGLMapSnapshotter alive through completion

1ec5:
Strongly capture MGLMapSnapshotter until its completion handler finishes executing. The developer no longer needs to explicitly capture the snapshotter in the completion handler to prevent it from being canceled or deallocated, because the internal completion handler does so for the developer.

Before #210, a snapshotter that wasn’t captured in its completion handler would be deallocated as soon as it went out of scope, but the completion handler would be called with an error “headlessly” after the snapshotter got deallocated. As of #210, the same snapshotter would’ve silently gone away without calling the completion handler, since it was considered the responsibility of the developer to hang onto the snapshotter. As of this PR, the developer is no longer required to hang onto the snapshotter, and the completion handler will generally receive a snapshot rather than an error.

This new behavior matches MapKit’s MKMapSnapshotter and is ideal for one-off usage. I had shied away from it in https://github.com/mapbox/mapbox-gl-native-ios/issues/200#issuecomment-598500827 because of the complexity around keeping the completion handler alive, but #210 gives us more confidence about the control flow and lifetime of any captured state. Nonetheless, this PR needs extra scrutiny because it possibly reintroduces edge cases that #210 had eliminated by reducing the snapshotter’s scope. In particular, we should make sure the application can reliably terminate while snapshotters are active without crashing.

Depends on #210. Fixes mapbox/mapbox-gl-native#12336 in a different way.

/cc @mapbox/maps-ios @alexshalamov

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -16,7 +16,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ### Snapshots
 
-* `MGLMapSnapshotter` no longer keeps itself from being deallocated before its completion handler is called. To ensure that the completion handler passed into `-[MGLMapSnapshotter startWithCompletionHandler:]` is called, maintain a strong reference to the snapshotter from a longer-lived class, such as the class where you call `-[MGLMapSnapshotter startWithCompletionHandler:]`. ([#210](https://github.com/mapbox/mapbox-gl-native-ios/pull/210))
+* You no longer need to explicitly capture the `MGLMapSnapshotter` object in the completion handler that you specify in `-[MGLMapSnapshotter startWithCompletionHandler:]`. Even if you declare the snapshotter locally without holding a strong reference to it, the snapshotter is only deallocated after the completion handler finishes, and the completion handler generally receives a valid snapshot. ([#210](https://github.com/mapbox/mapbox-gl-native-ios/pull/210))
 * The `-[MGLMapSnapshotter cancel]` method no longer calls the completion handler passed into `-[MGLMapSnapshotter startWithCompletionHandler:]`. ([#210](https://github.com/mapbox/mapbox-gl-native-ios/pull/210))
 * Fixed an issue where the `MGLMapSnapshotter.loading` property always returned `NO`, even while loading a snapshot. ([#210](https://github.com/mapbox/mapbox-gl-native-ios/pull/210))
 
--- a/platform/ios/Integration Tests/Snapshotter Tests/MGLMapSnapshotterTest.m
+++ b/platform/ios/Integration Tests/Snapshotter Tests/MGLMapSnapshotterTest.m
@@ -104,8 +104,7 @@ MGLMapSnapshotter* snapshotterWithBounds(MGLCoordinateBounds bounds, CGSize size
     CLLocationCoordinate2D coord   = CLLocationCoordinate2DMake(30.0, 30.0);
 
     // Test triggering to main queue
-    dispatch_queue_t backgroundQueue = dispatch_get_main_queue();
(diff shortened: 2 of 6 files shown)


platform/ios/CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ### Snapshots
 
-* `MGLMapSnapshotter` no longer keeps itself from being deallocated before its completion handler is called. To ensure that the completion handler passed into `-[MGLMapSnapshotter startWithCompletionHandler:]` is called, maintain a strong reference to the snapshotter from a longer-lived class, such as the class where you call `-[MGLMapSnapshotter startWithCompletionHandler:]`. ([#210](https://github.com/mapbox/mapbox-gl-native-ios/pull/210))
+* You no longer need to explicitly capture the `MGLMapSnapshotter` object in the completion handler that you specify in `-[MGLMapSnapshotter startWithCompletionHandler:]`. Even if you declare the snapshotter locally without holding a strong reference to it, the snapshotter is only deallocated after the completion handler finishes, and the completion handler generally receives a valid snapshot. ([#210](https://github.com/mapbox/mapbox-gl-native-ios/pull/210))
```
1ec5: Probably should link to this PR instead of #210.

platform/ios/Integration Tests/Snapshotter Tests/MGLMapSnapshotterTest.m
```diff
@@ -106,20 +108,29 @@ - (void)testDeallocatingSnapshotterDuringSnapshot🔒 {
...
-                __typeof__(self) strongself = weakself;
-                MGLTestFail(strongself, @"Completion handler should not be called after snapshotter is deallocated.");
-            }];
-            dispatch_group_leave(dg);
+            @autoreleasepool {
+                MGLMapSnapshotter *snapshotter = snapshotterWithCoordinates(coord, size);
+                weakSnapshotter = snapshotter;
+                
+                [snapshotter startWithCompletionHandler:^(MGLMapSnapshot * _Nullable snapshot, NSError * _Nullable error) {
+                    // We expect this completion block to be called with an error
```
1ec5: This comment is inaccurate; it crept back in when I reverted #210’s changes to the test, but we want the completion handler to be called with a valid snapshot now.

platform/ios/Integration Tests/Snapshotter Tests/MGLMapSnapshotterTest.m
```diff
@@ -106,20 +108,29 @@ - (void)testDeallocatingSnapshotterDuringSnapshot🔒 {
 //  dispatch_queue_t backgroundQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
 
     __weak __typeof__(self) weakself = self;
+    __block __weak MGLMapSnapshotter *weakSnapshotter;
 
     dispatch_async(backgroundQueue, ^{
```
julianrex: Although this is called `backgroundQueue` it's created with `dispatch_get_main_queue()` - is that WIP?

1ec5: I’m unsure. This test and the others introduced in mapbox/mapbox-gl-native#12355 all refer to `backgroundQueue`s that are actually the main queue.

1ec5: The tests do pass if I make them actual background queues though – would you prefer if I made that change here or keep the tests as-is for consistency?

julianrex: Yep, I think that was my original intention - I don't remember why it got commented out - I suspect that was during a test that was broken 🙁

PR #215, opened 2020-03-16
Fix UIColor alpha < 1 bug

jmkiley:
Addresses #125, but this time in the correct repo.

### DIFF
--- a/platform/darwin/src/MGLStyleValue_Private.h
+++ b/platform/darwin/src/MGLStyleValue_Private.h
@@ -214,7 +214,7 @@ private: // Private utilities for converting from mgl to mbgl values
 
     // Color
     void getMBGLValue(MGLColor *rawValue, mbgl::Color &mbglValue) {
-        mbglValue = rawValue.mgl_color;
+        mbglValue = rawValue.mgl_colorForPremultipliedValue;
     }
 
     // Image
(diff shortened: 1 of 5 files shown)


chloekraw:
Don't think we've got test coverage of this part of the code base to where we want it to be, and we should do that before merging a fix

jmkiley:
Makes sense! Will look into adding tests for this.

jmkiley:
The code here currently uses the approach from https://github.com/mapbox/mapbox-gl-native/pull/16153#issuecomment-580416201, but it doesn't seem to be executed with the `fillColor` property. Going to go back through and revisit @1ec5's suggestions from here https://github.com/mapbox/mapbox-gl-native/pull/16223#discussion_r383694089

nishant-karajgikar:
@jmkiley, what's the status of this PR?

jmkiley:
I put a `UIView` with an alpha of 0.5 in the upper left corner to compare with the `MGLFillStyleLayer` that is also set to a `UIColor` with an alpha of 0.5.
| Current behavior on master: | Behavior in https://github.com/mapbox/mapbox-gl-native-ios/pull/215/commits/9f353e21db37b79f151b7f512cf954dbe9d009c0 | Behavior with separate UIColor method |
---|----------|-----
|  |  |  |

nishant-karajgikar:
Closing in favor of #266

Issue #197, opened 2020-03-04
Cocoapods very long build time


arbitur:
>My setup
```
Machine: MacBook Pro 13", Mid 2014, 16GB RAM, i5 CPU
macOS version: 10.15.3
Mapbox version: 5.7.0, installed with cocoapods
XCode version: 11.3.1, using New build system
Cocoapods version: 1.9.0
```

Mapbox takes a tremendous amount of time during the "Embed Pods Frameworks" build process. On my machine (MacBook Pro 13", Mid 2014) it takes about **7 minutes** to compile and it stays on mapbox related frameworks solely for **~6 minutes**.

It didn't use to take this long like a year or so ago. I would guess this problem might have several causes, cocoapods version update, xcode version update, os update etc but I would like your input on this.

Do you have any tips on how to speed up this process? Thanks.

1ec5:
The only thing I can think of is that, beginning with version v5.6.0, the map SDK dynamically links MapboxMobileEvents and that library is built from source. Previously, it was statically linked into the map SDK, which is distributed as a binary.

/cc @rclee @julianrex

arbitur:
The fault wasnt with MapBox 👍

julianrex:
Thanks for letting us know @arbitur!

1ec5:
We’ve been encountering long MapboxMobileEvents link times on CircleCI that we suspect might be related to this issue. We aren’t sure if it always happens on CI, but it at least happens with the ios-release-tag builds that normally only run when releasing (the [worst possible time](https://en.wikipedia.org/wiki/Murphy's_law) for linking to stall).

@julianrex found that [reducing the number of concurrent jobs](https://app.circleci.com/pipelines/github/mapbox/mapbox-gl-native-ios/354/workflows/505a47d4-c095-4423-8373-887c6eb5dbc4/jobs/2739) seemed to reduce the link time, but we’d need to have the ios-release-tag workflow run to be sure. (#199 did temporarily enable the ios-release-tag workflow unconditionally, but I rewrote that out of history because we don’t want fake releases lying around.)

1ec5:
[This nightly build](https://circleci.com/gh/mapbox/mapbox-gl-native-ios/2905) also timed out due to long link times.

1ec5:
[...]

PR #221, opened 2020-03-17
Reduce MapboxMobileEvents link time

1ec5:
Limited the number of concurrent jobs when building the SDK to one less than the number of cores, in order to avoid the long link times seen most commonly in release builds.

Thanks to @julianrex for coming up with this potential fix for #197.

/cc @mapbox/maps-ios

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,8 @@ ifeq ($(shell uname -s), Darwin)
   HOST_PLATFORM = macos
   HOST_PLATFORM_VERSION = $(shell uname -m)
   export NINJA = platform/macos/ninja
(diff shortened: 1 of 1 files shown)


1ec5:
Looks like release builds are only [a minute shorter](https://app.circleci.com/pipelines/github/mapbox/mapbox-gl-native-ios/397/workflows/00ff932d-ab0f-4207-9e49-b8a27c63a733/jobs/2935) than [a build without this change](https://app.circleci.com/pipelines/github/mapbox/mapbox-gl-native-ios/396/workflows/d5432893-6f66-45d6-90cc-d75f2bc58f81/jobs/2929). 🤷‍♂

Issue #227, opened 2020-03-17
Uncaught exception starting MGLMapSnapshotter before using MGLMapView or MGLOfflineStorage


1ec5:
If an application attempts to take a snapshot before ever initializing an MGLMapView or interacting with MGLOfflineStorage directly, `-[MGLMapSnapshotter startWithQueue:overlayHandler:completionHandler:]` raises an exception:

https://github.com/mapbox/mapbox-gl-native-ios/blob/81df48df70ad6b116f3f17321d11a56a03b3cc3e/platform/darwin/src/MGLMapSnapshotter.mm#L267-L270

This check is premature, since the method soon afterwards causes MGLOfflineStorage to create an `mbgl::util::RunLoop` and various file sources on the current thread:

https://github.com/mapbox/mapbox-gl-native-ios/blob/81df48df70ad6b116f3f17321d11a56a03b3cc3e/platform/darwin/src/MGLMapSnapshotter.mm#L285 https://github.com/mapbox/mapbox-gl-native-ios/blob/81df48df70ad6b116f3f17321d11a56a03b3cc3e/platform/darwin/src/MGLMapSnapshotter.mm#L664 https://github.com/mapbox/mapbox-gl-native-ios/blob/81df48df70ad6b116f3f17321d11a56a03b3cc3e/platform/darwin/src/MGLOfflineStorage.mm#L64 https://github.com/mapbox/mapbox-gl-native-ios/blob/81df48df70ad6b116f3f17321d11a56a03b3cc3e/platform/darwin/src/MGLOfflineStorage.mm#L223 https://github.com/mapbox/mapbox-gl-native-ios/blob/81df48df70ad6b116f3f17321d11a56a03b3cc3e/platform/darwin/src/MGLFoundation.mm#L5

If we were to simply remove or defer this check, the check would pass even if starting the snapshotter on a background thread. However, the static `mbgl::util::RunLoop` intended for the main thread would be created for a background thread, which would lead to undefined behavior, for example when subsequently using MGLMapView.

Before #210, the same application didn’t trigger the exception because `-[MGLMapSnapshotter initWithOptions:]` happened to call `-setOptions:`, which called `+[MGLOfflineStorage sharedOfflineStorage]` before the run loop check. Even so, if a developer happened to initialize the snapshotter on a background thread, the static `RunLoop` intended for the main thread would’ve been created on the background thread.

[...]

PR #228, opened 2020-03-18
Adds NSURLSession delegate to MGLNetworkConfiguration

julianrex:
Depends on https://github.com/mapbox/mapbox-gl-native-ios/pull/242

This PR adds an undocumented `MGLNetworkConfigurationSessionDelegate` and a `id delegate` property to `MGLNetworkConfiguration`.

This allows the application to provide an `NSURLSession` object, which will be used instead of one being created using the existing `sessionConfiguration` property. It is the developer's responsibility to maintain the life (and reuse) the session object. It's also important to note that the delegate method does *not* get called on the main thread.

Background sessions are *not* supported, and an exception will fire if one is supplied - this exception is provided by the OS.

The existing `sessionConfiguration` behavior remains unchanged.

To enable these changes, the PR also includes:
- Removal of `MGLNetworkIntegrationManager` (non-public type) - this was just acting as a go-between, as was superfluous.
- Addition of `MGLIntegrationTestCase` to share existing testing functionality.

TODO:
- [x] Integration tests
- [x] update/test macOS

See also these [changes in gl-native](https://github.com/mapbox/mapbox-gl-native/tree/jrex/urlsession-poc), which this PR depends on.


julianrex:
cc @halset (who proposed a [session delegate PR](https://github.com/mapbox/mapbox-gl-native/pull/15128)).

halset:
@julianrex Thanks for the cc! Should probably recreate the [`NSURLSessionDelegate`-PR](https://github.com/mapbox/mapbox-gl-native/pull/15128) after this is merged to be able to set a `NSURLSessionDelegate` or will this PR handle that as well?

julianrex:
@halset I'm thinking this PR would make the existing session delegate PR redundant.

Note this is still just an experiment, and we may end up adding a delegate for `MGLNetworkConfiguration` which I think would be more idiomatic.

platform/darwin/src/MGLNetworkConfiguration.m
```diff
@@ -1,21 +1,28 @@
...
 
 static NSString * const MGLStartTime = @"start_time";
 static NSString * const MGLResourceType = @"resource_type";
 NSString * const kMGLDownloadPerformanceEvent = @"mobile.performance_trace";
 
-@interface MGLNetworkConfiguration ()
+@interface MGLNetworkConfiguration () <MGLNativeNetworkDelegate>
 
-@property (strong) NSURLSessionConfiguration *sessionConfig;
+//@property (strong) NSURLSessionConfiguration *sessionConfig;
```
fabian-guerra: nit: please remove unused code.

platform/darwin/src/MGLNetworkConfiguration.m
```diff
@@ -34,7 +41,8 @@ + (instancetype)sharedManager {
         _sharedManager = [[self alloc] init];
     });
 
-    [self setNativeNetworkManagerDelegateToDefault];
+    // Notice, this is reset for each call.
+    [MGLNativeNetworkManager sharedManager].delegate = _sharedManager;
```
fabian-guerra: What is the purpose of resetting delegate property? Maybe you can point to code that creates a new sharedManager?

julianrex: Added additional comments.

platform/darwin/darwin.xcconfig
```diff
@@ -56,7 +56,19 @@ GCC_WARN_UNUSED_FUNCTION = YES
...
+HEADER_SEARCH_PATHS = $(HEADER_SEARCH_PATHS) "$(PROJECT_DIR)/../../vendor/mapbox-gl-native/vendor/mapbox-base/mapbox/optional"
+HEADER_SEARCH_PATHS = $(HEADER_SEARCH_PATHS) "$(PROJECT_DIR)/../../vendor/mapbox-gl-native/vendor/mapbox-base/mapbox/variant/include"
+HEADER_SEARCH_PATHS = $(HEADER_SEARCH_PATHS) "$(PROJECT_DIR)/../../vendor/mapbox-gl-native/vendor/mapbox-base/extras/expected-lite/include"
+HEADER_SEARCH_PATHS = $(HEADER_SEARCH_PATHS) "$(PROJECT_DIR)/../../vendor/mapbox-gl-native/vendor/mapbox-base/extras/rapidjson/include"
+HEADER_SEARCH_PATHS = $(HEADER_SEARCH_PATHS) "$(PROJECT_DIR)/../../vendor/mapbox-gl-native/vendor/polylabel/include"
+HEADER_SEARCH_PATHS = $(HEADER_SEARCH_PATHS) "$(PROJECT_DIR)/../../platform/darwin/include"
+HEADER_SEARCH_PATHS = $(HEADER_SEARCH_PATHS) "$(PROJECT_DIR)/../../vendor/mapbox-gl-native/include"
+HEADER_SEARCH_PATHS = $(HEADER_SEARCH_PATHS) "$(PROJECT_DIR)/../../vendor/mapbox-gl-native/platform/default/include"
+HEADER_SEARCH_PATHS = $(HEADER_SEARCH_PATHS) "$(PROJECT_DIR)/../../vendor/mapbox-gl-native/platform/darwin/include"
```
1ec5: I think .xcconfig is more like a property list than a script. If you specify `HEADER_SEARCH_PATHS` multiple times, each new specification overrides the previous one. I think that’s why the builds are all failing with stuff like:

```
❌  /Users/distiller/project/vendor/mapbox-gl-native/include/mbgl/util/geometry.hpp:3:10: 'mapbox/geometry/geometry.hpp' file not found



#include <mapbox/geometry/geometry.hpp>

         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

platform/ios/Integration Tests/MGLNetworkConfigurationIntegrationTests.mm
```diff
@@ -47,38 +54,214 @@ - (void)test1_NativeNetworkManagerDelegateIsSet
...
+    ASSERT_NATIVE_DELEGATE_IS_NIL();
     (void)[[MGLMapView alloc] init];
-    ASSERT_DELEGATE_IS_NOT_NIL();
+    ASSERT_NATIVE_DELEGATE_IS_NOT_NIL();
+}
+
+- (void)testNetworkConfigurationDelegateIsNil
+{
+    MGLNetworkConfiguration *manager = [MGLNetworkConfiguration sharedManager];
+    XCTAssertNil(manager.delegate);
```
1ec5: Would it be possible to convert these integration tests into unit tests? At a glance, they don’t look too different than the existing unit tests of MGLOfflineStorage.

platform/macos/src/MGLMapView.mm
```diff
@@ -259,7 +259,8 @@ + (NSArray *)restorableStateKeyPaths {
 }
 
 - (void)commonInit {
-    MGLNativeNetworkManager.sharedManager.delegate = MGLNetworkIntegrationManager.sharedManager;
+    [MGLNetworkConfiguration setNativeNetworkManagerDelegateToDefault];
```
1ec5: Incidentally, this is an odd name for a non-setter. `-resetNativeNetworkManagerDelegate` would be more intuitive.

platform/darwin/darwin.xcconfig
julianrex: This is already done.

platform/macos/src/MGLMapView.mm
julianrex: I like `reset...`. Have included it in a recent push.

platform/ios/Integration Tests/MGLNetworkConfigurationIntegrationTests.mm
julianrex: Yes, I think so; let's do that in another PR.

platform/darwin/src/MGLNetworkConfiguration.h
```diff
@@ -4,13 +4,39 @@
...
  `NSURLSessionConfiguration`, and other resources.
  */
 MGL_EXPORT
 @interface MGLNetworkConfiguration : NSObject
 
+/**
+ :nodoc:
+ Delegate for the `MGLNetworkConfiguration` class.
+ */
+@property (nonatomic, weak) id delegate;
```
1ec5:
```suggestion
@property (nonatomic, weak) id <MGLNetworkConfigurationSessionDelegate> delegate;
```

platform/darwin/src/MGLNetworkConfiguration.m
```diff
@@ -34,26 +40,46 @@ + (instancetype)sharedManager {
         _sharedManager = [[self alloc] init];
     });
 
-    [self setNativeNetworkManagerDelegateToDefault];
+    // Notice, this is reset for each call. This is primarily for testing purposes.
+    // TODO: Consider only calling this for testing?
```
1ec5: Is this to-do still relevant? (×2)

platform/darwin/src/MGLNetworkConfiguration.h
```diff
@@ -4,13 +4,39 @@
...
+/**
+ :nodoc:
+ The `MGLNetworkConfigurationSessionDelegate` protocol provides an `NSURLSession`
+ object for the `MGLNetworkConfiguration`. This API should be considered
+ experimental, likely to be removed or changed in future releases.
+
+ This method is called from background threads, i.e. it is not called on the main
+ thread.
+
+ @note: Background sessions (i.e. created with
```
1ec5:
```suggestion
 @note Background sessions (i.e. created with
```

platform/darwin/src/MGLNetworkConfiguration.h
```diff
@@ -23,12 +49,14 @@ MGL_EXPORT
...
- Assign this object before instantiating any `MGLMapView` object.
+ Assign this object before instantiating any `MGLMapView` object, or using
+ `MGLOfflineStorage`
  
  @note: `NSURLSession` objects store a copy of this configuration. Any further changes
  to mutable properties on this configuration object passed to a session’s initializer
  will not affect the behavior of that session.
- 
+
+ @note: Background sessions are not currently supported.
```
1ec5:
```suggestion
 @note Background sessions are not currently supported.
```

platform/darwin/src/MGLNetworkConfiguration.h
```diff
@@ -23,12 +49,14 @@ MGL_EXPORT
  If this property is set to nil or if no session configuration is provided this property
  is set to the default session configuration.
  
- Assign this object before instantiating any `MGLMapView` object.
+ Assign this object before instantiating any `MGLMapView` object, or using
+ `MGLOfflineStorage`
```
1ec5: What’s an example of a place where this object can be assigned safely? Consider that fixing #227 may require us to invoke `+[MGLOfflineStorage sharedStorage]` in an `+initialize` method.

platform/darwin/src/MGLNetworkConfiguration.h
julianrex: This is deliberately left as `id` for a couple of reasons:
- to keep this clearly undocumented
- so that delegates can optionally conform to `MGLNetworkConfigurationSessionDelegate` (rather than using `@optional` methods and hitting Swift issues). Perhaps there's a better way here?

platform/darwin/src/MGLNetworkConfiguration.m
julianrex: I'll create a new ticket to track this.

platform/darwin/src/MGLNetworkConfiguration.h
julianrex: Ugh. Let's address that when we fix #227. Hopefully we can find a way around using an `+initialize`.

julianrex:
When merged, the commit will need to be cherry-picked to `release-vanillashake`

platform/darwin/src/MGLNetworkConfiguration.h
1ec5: > to keep this clearly undocumented

Neither `delegate` nor `MGLNetworkConfigurationSessionDelegate` will become documented by virtue of using a strong type. On the other hand, we lose helpful compiler warnings when we misuse `delegate` internally.

> so that delegates can optionally conform to `MGLNetworkConfigurationSessionDelegate` (rather than using `@optional` methods and hitting Swift issues). Perhaps there's a better way here?

What Swift issues are you referring to? Swift supports optional methods in Objective-C protocols. We should make this protocol’s methods optional and require the delegate to be of the right type.

julianrex: > What Swift issues are you referring to?

Ah, I was thinking of the problems with optional properties & Swift. I've updated the type, and marked the method as optional.

PR #231, opened 2020-03-19
Refactor metrics triggering

julianrex:
Rejigs metrics triggering so that it can be tested locally (to avoid repeated PRs & failures!), for example:
```
curl -u <my-token>: -d build_parameters[CIRCLE_JOB]=ios-trigger-metrics https://circleci.com/api/v1.1/project/github/mapbox/mapbox-gl-native-ios/tree/jrex/update-metrics-trigger-fix2
```

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -216,7 +216,7 @@ commands:
           name: Trigger metrics
           command: |
             if [ -n "${MOBILE_METRICS_TOKEN}" ]; then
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -419,16 +436,6 @@ jobs:
...
-      - run:
-          name: Trigger metrics
-          command: |
-            if [[ $CIRCLE_BRANCH == master ]]; then
-              if [ -n "${MOBILE_METRICS_TOKEN}" ]; then
-                curl -u ${CIRCLE_CI_TOKEN}: -X POST --header "Content-Type: application/json" --data '{"parameters": {"run_ios_maps_benchmark": true, "ci_ref" : $CIRCLE_BUILD_NUM }, "branch": "master"}' https://circleci.com/api/v2/project/github/mapbox/mobile-metrics/pipeline
-              else
-                echo "MOBILE_METRICS_TOKEN not provided"
-              fi
-            fi
```
jmkiley: Should we break this out into a separate script? Not sure how much we want to do in the config.yml itself.

PR #235, opened 2020-03-23
Snapshotter (runtime styling) delegates

### LINKED ISSUE #200: Platform bindings for runtime styling APIs on `MapSnapshotter`

### LINKED ISSUE #236: Snapshotter overlay handler not called
The overlay handler passed into `-[MGLMapSnapshotter startWithOverlayHandler:completionHandler:]` is never called on macOS as of v0.15.0-beta.1.

This regression was introduced in #210. `MGLAttributedSnapshot()` calls `overlayHandler()` on iOS but not on macOS.

1ec5:
Added a protocol and property for MGLMapSnapshotter’s delegate, which it uses for runtime styling. Generalized MGLStyle to be usable in the absence of MGLMapView. Moved MGLOpenGLStyleLayer code around to compile.

In macosapp, the File ‣ Export Image command now matches the main map’s layer visibility settings and translated labels.

Added unit tests and snapshot tests of MGLMapSnapshotter that work on both iOS and macOS, unlike the iOS-only integration tests of MGLMapSnapshotter. This test suite comes with an easy way to snapshot-test runtime styling.

Fixes #200, fixes #236.

/cc @mapbox/maps-ios


platform/darwin/src/MGLMapSnapshotter.h
```diff
@@ -284,6 +288,78 @@ MGL_EXPORT
...
+ snapshotting the map.
+ 
+ This method may be called for a variety of reasons, including a network
+ connection failure or a failure to fetch the style from the server. You can use
+ the given error message to notify the user that map data is unavailable.
+ 
+ @param snapshotter The snapshotter that is unable to load the data.
+ @param error The reason the data could not be loaded.
+*/
+- (void)mapSnapshotterDidFail:(MGLMapSnapshotter *)snapshotter withError:(NSError *)error;
```
julianrex: Does this overlap with the error code passed to the `MGLMapSnapshotCompletionHandler`? I think there might be some scope for confusion here.

platform/darwin/src/MGLMapSnapshotter.h
```diff
@@ -284,6 +288,78 @@ MGL_EXPORT
...
+ `-[MGLMapSnapshotter startWithQueue:completionHandler:]` as long as the
+ `MGLMapSnapshotter.delegate` property is set. Changes to sources or layers of
+ the style being snapshotted do not cause this method to be called.
+ 
+ @param snapshotter The snapshotter that has just loaded a style.
+ @param style The style that was loaded.
+ */
+- (void)mapSnapshotter:(MGLMapSnapshotter *)snapshotter didFinishLoadingStyle:(MGLStyle *)style;
+
+- (void)mapSnapshotter:(MGLMapSnapshotter *)snapshotter didFailLoadingImageNamed:(NSString *)name;
```
julianrex: If we choose not to support this delegate method, should we move the `mapSnapshotter:didFinishLoadingStyle:` to another block handler? Alternatively, if we feel using a delegate is preferable how about we add a finish method and deprecate the handlers? (I think you discussed this before, so apologies if I'm just catching up).

1ec5:
> Also, do you have any tests for the new style handling in snapshotter?

Just the manual macosapp thing so far, but I can add an integration test.

platform/darwin/src/MGLMapSnapshotter.h
1ec5: So far this delegate method only gets called if the style fails to load. But I intentionally worded it more generically in case we need to support other failures in the future this way, particularly recoverable errors. A failure to load the style also causes the snapshot to fail, so an error still gets passed into the completion handler. This delegate method just gives additional information about the cause.

platform/darwin/src/MGLMapSnapshotter.h
1ec5: I went back and forth on whether to introduce an additional block parameter as a convenience to avoid having to write a delegate. I would’ve liked to use the “action block” pattern in https://github.com/mapbox/mapbox-gl-native-ios/issues/200#issuecomment-598464934. The problem is that there are already two block parameters that get called on different queues, so adding a third results in a pretty complicated call site plus confusion over which blocks get called on which queues.

platform/darwin/src/MGLMapSnapshotter.h
julianrex: So to confirm - in the case of an error where the dev implements this method - they would see the error here, and then finally in the completion block? Would it be the same error or this one would be more specific?

platform/darwin/src/MGLMapSnapshotter.h
julianrex: 👍 Let's consider deprecating the overlay handler in favour of a delegate method in a future release?

1ec5: The difference is that the delegate methods are all called on the same thread that the snapshot was started on. Typically this would be the main thread. (This is how mapbox/mapbox-gl-native#16268 was designed.) Based on the discussion in https://github.com/mapbox/mapbox-gl-native-ios/pull/210#discussion_r392107428, it’s unlikely that we’d move the overlay handler to the main thread. I’m wary of encouraging the developer to write a class whose methods are called on different threads, which would require internal synchronization that even the best developers get wrong sometimes.

platform/darwin/src/MGLMapSnapshotter.h
1ec5: Yes, they’d see both errors. This error would be more specific, though I wouldn’t make that part of the API contract since mbgl’s API contract doesn’t say what errors show up when.

platform/darwin/src/MGLOpenGLStyleLayer.mm
```diff
@@ -107,13 +68,20 @@ - (instancetype)initWithIdentifier:(NSString *)identifier {
     return (mbgl::style::CustomLayer *)super.rawLayer;
 }
 
+- (MGLMapView *)mapView {
+    if ([self.style.stylable isKindOfClass:[MGLMapView class]]) {
```
1ec5: As implemented in this PR, MGLOpenGLStyleLayer doesn’t support snapshotting, but it wouldn’t be too difficult to make it work if necessary.

1ec5:
I’ve added unit tests and snapshot tests of MGLMapSnapshotter that work on both iOS and macOS, unlike the iOS-only integration tests of MGLMapSnapshotter, which will help catch issues like #236. This test suite comes with an easy way to snapshot-test runtime styling for other purposes too.

PR #252, opened 2020-04-01
Fix crash when quitting while a map view has annotations

### LINKED ISSUE #87: MGLUnderlyingMapUnavailableException - MGLMapView The underlying map is not available - this happens during app termination
Hey there,

our crash reporting tool regularly reports the following exception:

`MGLUnderlyingMapUnavailableException · The underlying map is not available - this happens during app termination`

Unfortunately I am not able to reproduce the behavior, so besides the description of the exception, I have no idea why this is happening and what kind of impact the exception has.

Can someone help?

### Steps to reproduce

 1.
 2.
 3.

### Configuration

**Mapbox SDK versions:** Mapbox-iOS-SDK (5.3.0)
**iOS/macOS versions:** iOS 12 and 13
**Device/simulator models:** iPhone 6S for example
**Xcode version:** 11

1ec5:
If the underlying `mbgl::Map` has been torn down, so has the mbgl annotation, so don’t worry about the `mbgl::Map`, but keep unregistering KVO and other SDK-side annotation resources.

Also wrapped `MGLMapView`’s `mbgl::Map` reference in a `std::unique_ptr` for increased pointer safety.

Fixes #87.

/cc @mapbox/maps-ios

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -30,6 +30,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 ### Other changes
 
 * Added the `MGLMapView.minimumPitch` and `MGLMapView.maximumPitch` properties to further limit how much the user or your code can tilt the map. ([#208](https://github.com/mapbox/mapbox-gl-native-ios/pull/208))
+* Fixed a crash while quitting the application after adding an annotation to `MGLMapView`. ([#252](https://github.com/mapbox/mapbox-gl-native-ios/pull/252))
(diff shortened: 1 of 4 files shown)


chloekraw:
@1ec5 let's cherrypick this into the 5.8.0 release?

1ec5:
master and release-vanillashake have been even with each other this release cycle, so I’ve just been fast-forward-merging master into release-vanillashake without any cherry-picking.

nishant-karajgikar:
Thanks for this @1ec5 !

Issue #154, opened 2020-01-25
The recent requirement of "use_frameworks!"


kristfal:
This is a continuation of the closed topic at #131, feel free to close this issue and reopen that if it makes more sense.

5.6.0 introduced a hard requirement for pod installation to enable `use_frameworks!`. 

While this works well for this library in isolation, a fair share of other iOS libraries fails to compile with dynamic frameworks, including Firebase, Google Analytics++. 

For any app with dependencies like the ones mentioned above, it seems like the only option is to drop the Mapbox pods and embed pre-downloaded binaries. This is a poor developer experience for many reasons. 

For anyone maintaining downstream projects, this means having to pick between asking users to enable `use_frameworks!` or guide them through downloading and installing static frameworks. Both options are far from good, and Carthage is not an option for any ReactNative based project. 

With this in mind, how do you recommend we handle this situation? Would you be able to reverse the use_frameworks requirement?

Some relevant downstream tickets: 
https://github.com/mapbox/mapbox-navigation-ios/issues/2309
https://github.com/react-native-mapbox-gl/maps/issues/635

chrisballinger:
I came here to raise the same issue. Our app uses static linking for launch performance reasons, with `use_modular_headers!` in the `Podfile` instead of `use_frameworks!`. 

Additionally [CocoaPods 1.9](http://blog.cocoapods.org/CocoaPods-1.9.0-beta/) is introducing a static framework linking option `use_frameworks! :linkage => :static`, which will trigger the same issue. [This feature](https://github.com/CocoaPods/CocoaPods/issues/9099) will eventually allow individual pods to be marked as statically or dynamically linked, but in the meantime we will be forced to use a workaround.

Another solution is that Mapbox (or the community) could ship another podspec with a statically linked framework that uses [`spec.static_framework = true`](https://guides.cocoapods.org/syntax/podspec.html#static_framework).

julianrex:
Thanks both of you for the detailed comments - we are currently evaluating the most appropriate path forwards.

> Would you be able to reverse the use_frameworks requirement?

This is part of our discussions; we also need to consider our other distribution methods and ensure that we're consistent across the board.

@chrisballinger thanks for the link about 1.9. 

Providing another podspec is another thing we're considering. We will update this thread when there's more clarity.

kristfal:
Thanks for the update. I know this is a complex topic, but any idea when you plan to make a decision? 

[...]

PR #267, opened 2020-04-10
Provide Static subspec, for using without use_frameworks!

mfazekas:
`<changelog>Mapbox-iOS-SDK pod now can be consumed without use_frameworks! using the Mapbox-iOS-SDK/Static subspec.</changelog>`

Fixes: #154 

```ruby
pod 'Mapbox-iOS-SDK/Static', '~>5.8'
```

or

```ruby
use_frameworks!
pod 'Mapbox-iOS-SDK', '~>5.8'
```

Btw 'Mapbox-iOS-SDK/Static' works with `use_frameworks!` too, but not vice versa.


stonetip:
Is the "use_frameworks!" issue ever going to be resolved, particularly in July when the new Mapbox GL Native renderer is introduced? (Kinda hoping to upgrade for that).

PR #268, opened 2020-04-14
[ios] add allowScrollGesturesDuringRotateOrZoom

iBinh:
Added new property `MGLMapView.allowScrollGesturesDuringRotateOrZoom`, which allows you to change the center coordinate of the map while rotating or zooming

### DIFF
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -741,6 +741,10 @@ MGL_EXPORT
 @property(nonatomic, getter=isPitchEnabled) BOOL pitchEnabled;
 
 /**
+ A Boolean value that determines whether the user can change the center coordinate of the map while rotating or zooming. When this property is set to YES, the default, the location of the gesture in the map view determines the map's center coordinate.
+ */
+@property(nonatomic) BOOL allowScrollGesturesDuringRotateOrZoom;
+/**
  A Boolean value that determines whether the user will receive haptic feedback
  for certain interactions with the map.
 
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -741,9 +741,10 @@ MGL_EXPORT
 @property(nonatomic, getter=isPitchEnabled) BOOL pitchEnabled;
 
 /**
- A Boolean value that determines whether the user can change the center coordinate of the map while rotating or zooming. When this property is set to YES, the default, the location of the gesture in the map view determines the map's center coordinate.
+ A Boolean value that determines whether gestures are anchored to the center coordinate of the map while rotating or zooming.
+ Default value is set to NO.
  */
-@property(nonatomic) BOOL allowScrollGesturesDuringRotateOrZoom;
+@property(nonatomic) BOOL anchorRotateOrZoomGesturesToCenterCoordinate;
 /**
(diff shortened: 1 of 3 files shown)


platform/ios/src/MGLMapView.h
```diff
@@ -740,6 +740,10 @@ MGL_EXPORT
  */
 @property(nonatomic, getter=isPitchEnabled) BOOL pitchEnabled;
 
+/**
+ A Boolean value that determines whether the user can change the center coordinate of the map while rotating or zooming. When this property is set to YES, the default, the location of the gesture in the map view determines the map's center coordinate.
+ */
+@property(nonatomic) BOOL allowScrollGesturesDuringRotateOrZoom;
```
fabian-guerra, line 746: I think the naming of this property does not reflect the real purpose. I was thinking if we could change this to `isCenterCoordinateGestureAnchor` with a default value = `NO`

platform/ios/src/MGLMapView.mm
```diff
@@ -1714,6 +1717,12 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
 {
     if ( ! self.isScrollEnabled) return;
 
+    if (!self.allowScrollGesturesDuringRotateOrZoom) {
```
fabian-guerra, line 1720: Is it possible to move this code to `gestureRecognizerShouldBegin`?

platform/ios/src/MGLMapView.h
iBinh: This name is same as Google Maps, I think we don't need to change it

platform/ios/src/MGLMapView.mm
iBinh: Yes

platform/ios/src/MGLMapView.h
fabian-guerra: Hi. Thank you for your reply. Although Google is using something similar is also misleading. Perhaps:
```
anchorRoteOrZoonGesturesToCenterCoordinate
```
Communicates better that when turned on it will use center coordinate as anchor.

iBinh: Did you mean `anchorRotateOrZoomGesturesToCenterCoordinate` ?
It sounds better!

fabian-guerra: Hi. Yes, I mean that. Fixing my typos.

iBinh: 😬  Do you have comment about it's description?

fabian-guerra: > A Boolean value that determines whether gestures are anchored to the center coordinate of the map while rotating or zooming. 
Default value is set to `NO`.

iBinh: fixed

iBinh:
> One final task. Could you please add this blurb to the changelog.
> 
> > * Added `MGLMapView.anchorRotateOrZoomGesturesToCenterCoordinate ` property that anchors zoom and rotating gestures to the map's view center coordinate. ([#268](https://github.com/mapbox/mapbox-gl-native-ios/pull/268))

added

captainbarbosa:
Just ran this on our CI - thanks for your contribution @iBinh!

PR #280, opened 2020-04-24
Snapshot style’s default camera unless overridden

### LINKED ISSUE #279: MGLMapSnapshotter is centered at Null Island instead of the style’s default camera
If a style specifies its initial center coordinate and zoom level via the `center` and `zoom` properties, respectively, and MGLMapSnapshotter is created with a completely unset camera, the resulting snapshot should be looking at the style’s initial center coordinate and zoom level. Instead, it is centered at Null Island.

To reproduce this issue:

```objc
MGLMapCamera *camera = [MGLMapCamera camera];
camera.centerCoordinate = kCLLocationCoordinate2DInvalid;
camera.heading = -1;
camera.pitch = -1;
MGLMapSnapshotOptions *options = [[MGLMapSnapshotOptions alloc] initWithStyleURL:styleURL camera:camera size:expectedImage.size];
options.zoomLevel = -1;

MGLMapSnapshotter *snapshotter = [[MGLMapSnapshotter alloc] initWithOptions:options];
snapshotter startWithCompletionHandler:^(MGLMapSnapshot * _Nullable snapshot, NSError * _Nullable error) {
    // Examine the image.
}];
```

The iOS/macOS snapshotter’s behavior differs from, say, `mbgl::HeadlessFrontend` in `MapMode::Static`.

/cc @mapbox/maps-ios

1ec5:
Only set a snapshotter’s underlying camera options – overriding the style’s default – if the application has set a valid `MGLMapCamera.centerCoordinate`, nonnegative `MGLMapCamera.heading`, nonnegative `MGLMapCamera.pitch`, or nonnegative `MGLMapSnapshotOptions.zoomLevel`.

Fixes #279.

/cc @mapbox/maps-ios

### DIFF
--- a/platform/darwin/src/MGLMapSnapshotter.mm
+++ b/platform/darwin/src/MGLMapSnapshotter.mm
@@ -736,10 +736,18 @@ NSArray<MGLAttributionInfo *> *MGLAttributionInfosFromAttributions(mbgl::MapSnap
     if (CLLocationCoordinate2DIsValid(options.camera.centerCoordinate)) {
         cameraOptions.center = MGLLatLngFromLocationCoordinate2D(options.camera.centerCoordinate);
     }
-    cameraOptions.bearing = MAX(0, options.camera.heading);
-    cameraOptions.zoom = MAX(0, options.zoomLevel);
-    cameraOptions.pitch = MAX(0, options.camera.pitch);
-    _mbglMapSnapshotter->setCameraOptions(cameraOptions);
+    if (options.camera.heading >= 0) {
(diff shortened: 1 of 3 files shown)


1ec5:
The risk from this change is that someone may have been relying on being able to set the heading to a negative value to rotate the snapshot counterclockwise from north. That would be a misuse of [`CLLocationDirection`](https://developer.apple.com/documentation/corelocation/cllocationdirection) according to Apple documentation, but it’s an easy enough mistake to make in Objective-C.

PR #289, opened 2020-04-30
Add offline pack context setter

### LINKED ISSUE #256: Allow changing offline metadata
`mbgl::OfflineRegion` supports getting the region’s metadata but not setting it. An offline region’s metadata typically contains a name. An application that allows the user to choose the name would typically also allow the user to change the name.

/cc @jfirebaugh @zugaldia

1ec5:
Added the `-[MGLOfflinePack setContext:completionHandler:]` method for replacing the data associated with an offline pack, such as a name.

To do:

* [x] Add setter
* [x] Add unit test of setter
* [x] Debug unit test failure
* [ ] Make offline pack table editable in macosapp

Fixes #256.

/cc @mapbox/maps-ios

### DIFF
--- a/platform/darwin/test/MGLOfflineStorageTests.mm
+++ b/platform/darwin/test/MGLOfflineStorageTests.mm
@@ -139,6 +139,22 @@
     }];
     [pack requestProgress];
     [self waitForExpectationsWithTimeout:5 handler:nil];
+    
+    XCTAssertEqualObjects(pack.context, context, @"Offline pack context should match the context specified by the application.");
+    NSString *newName = @"🍑 Peach Grove";
+    NSData *newContext = [NSKeyedArchiver archivedDataWithRootObject:@{
+        nameKey: newName,
+    }];
+    
+    XCTestExpectation *contextCompletionHandlerExpectation = [self expectationWithDescription:@"set pack completion context handler"];
+    __weak MGLOfflinePack *weakPack = pack;
+    [pack setContext:newContext completionHandler:^(NSError * _Nullable error) {
+        XCTAssertNil(error);
+        XCTAssertNotNil(weakPack);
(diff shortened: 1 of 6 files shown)


platform/darwin/test/MGLOfflineStorageTests.mm
```diff
@@ -139,6 +139,22 @@ - (void)testAddPackForBounds {
...
+    NSData *newContext = [NSKeyedArchiver archivedDataWithRootObject:@{
+        nameKey: newName,
+    }];
+    
+    XCTestExpectation *contextCompletionHandlerExpectation = [self expectationWithDescription:@"set pack completion context handler"];
+    __weak MGLOfflinePack *weakPack = pack;
+    [pack setContext:newContext completionHandler:^(NSError * _Nullable error) {
+        XCTAssertNil(error);
+        XCTAssertNotNil(weakPack);
+        XCTAssertEqualObjects(weakPack.context, newContext, @"Offline pack context should match the updated context specified by the application.");
```
1ec5, line 154: This assertion currently fails every time due to mapbox/mapbox-gl-native#16450. Until that issue is fixed, the call to `-[MGLOfflinePack setContext:completionHandler:]` will appear to have failed silently, even if it did result in a successful change to the database.

The developer could work around this issue by calling `-[MGLOfflineStorage reloadPacks]` in the completion handler, but that has the side effect of invalidating any MGLOfflinePack object the application is currently holding onto.

1ec5:
8e5dd29ff47e6f75607f2353c9d0703322c94381 tries to work around mapbox/mapbox-gl-native#16450 by reloading the offline packs, storing the `mbgl::OfflineRegion` matching the offline pack being updated, and throwing away all the wrapping MGLOfflinePacks. Unfortunately, calling `mbgl::DatabaseFileSource::listOfflineRegions()` seems to return exactly the same offline regions as before, making me wonder if mbgl is actually changing anything like it’s supposed to.

PR #189, opened 2020-03-01
Allow specifying multiple fonts or font families for local font rendering

1ec5:
Removed `MGLIdeographicFontFamilyName` validation from MGLRendererConfiguration, since `mbgl::LocalGlyphRasterizer` performs more robust validation. Removed C++ from MGLRendererConfiguration interface, implementation, and tests and simplified the logic.

Pass all the specified font and font family names into mbgl so that `mbgl::LocalGlyphRasterizer` can form a full font cascade list.

* [x] Plumb through font name list
* [x] Clean up MGLRendererConfiguration
* [x] Remove font family validation from MGLRendererConfiguration
* [x] Clean up `MGLIdeographicFontFamilyName` documentation
* [ ] Plumb through an array of strings instead of a line-delimited string
* [x] Update submodule

Fixes #105. Depends on #280, mapbox/mapbox-gl-native#16253, and mapbox/mapbox-gl-native#16428.

/cc @mapbox/maps-ios

PR #292, opened 2020-04-30
Add guide about customizing fonts

1ec5:
Split out a guide about customizing fonts, covering both server-side and client-side fonts following the changes in #189.

customizing-fonts

/cc @mapbox/maps-ios @mapbox/docs (FYI)


platform/darwin/docs/guides/Customizing Fonts.md
```diff
@@ -0,0 +1,21 @@
...
+
+If none of the symbol layer’s fonts is available, or if these fonts do not contain the required glyphs, then the map tries to find a matching font in the [`MGLIdeographicFontFamilyName`](infoplist-keys.html#mglideographicfontfamilyname) Info.plist key. Like the `MGLSymbolStyleLayer.textFontNames` property, this key can contain a family name, display name, or PostScript name. This key is a global fallback that applies to all layers uniformly. It can either be a single string or an array of strings, which the map tries to apply in order from most preferred to least preferred.
```
chloekraw, line 17:
```suggestion
If none of the fonts specified on a symbol layer are available, or if the fonts do not contain the required glyphs, then the map tries to find a matching font in the [`MGLIdeographicFontFamilyName`](infoplist-keys.html#mglideographicfontfamilyname) Info.plist key. Like the `MGLSymbolStyleLayer.textFontNames` property, this key can contain a family name, display name, or PostScript name. This key is a global fallback that applies to all layers uniformly. It can either be a single string or an array of strings, which the map tries to apply in order from most preferred to least preferred.
```

chloekraw: Minor nitpicks:
- "none" should be followed by a plural verb because in this context, none refers to "if not any of the fonts are available". None is singular when it is used in reference to part of a single item, e.g., "none of my bread is yours."
- suggested a rewording of the first clause of the first sentence that I think is clearer

chloekraw, line 17: One thought I'm mulling over: instead of writing out the information about `MGLIdeographicFontFamilyName` here, what do you think about linking to that section of the Info.plist guide for the details? I am generally hesitant to repeat information in more than one place, because it's one more place to remember to update if the behavior changes, which increases the chance that the information will go stale.

That said, at the same time, I really like the way this documentation reads as-is, so I don't have strong feelings about this.

1ec5: This is an area where [style guides differ](https://www.dictionary.com/e/none/). I’ve always followed the “not one” rule of thumb from [a grade school English textbook](http://thewritesource.com/), but it’s admittedly quite pedantic and it looks like some style guides like APA have gone the other way. Our internal style guide is silent on the matter.

1ec5: The guide does link to the relevant section of the Info.plist documentation, but the idea was to draw a parallel between the two places that accept font names. Fortunately, since these details are now part of the public API, it’s unlikely that they’ll change in a minor release.

PR #296, opened 2020-05-01
Mapbox Maps SDK for iOS v5.9.0-beta.1, macOS v0.16.0-beta.1

### LINKED ISSUE #262: Adopt Apple Generic Versioning
The iOS and macOS Xcode projects should adopt the [Apple Generic Versioning](https://developer.apple.com/library/archive/qa/qa1827/_index.html) system to replace the incomplete, ad-hoc system we’re currently using.

## Background

Frameworks and applications are supposed to set the `CURRENT_PROJECT_VERSION` and `CURRENT_SHORT_VERSION` build settings to something meaningful. These settings are used in several ways, such as to display the application version in Finder (handy for debugging purposes) or to query a framework’s version at runtime (an additional layer of safety when weak-linking a framework).

We currently rely on the two package.sh scripts to fill in the iOS and macOS map SDKs’ `CURRENT_PROJECT_VERSION` and `CURRENT_SHORT_VERSION` build settings:

https://github.com/mapbox/mapbox-gl-native-ios/blob/69ddd885126a9650dbe1d20f4b7cee73aa7fc946/platform/ios/scripts/package.sh#L68-L69 https://github.com/mapbox/mapbox-gl-native-ios/blob/69ddd885126a9650dbe1d20f4b7cee73aa7fc946/platform/ios/scripts/package.sh#L88-L89 https://github.com/mapbox/mapbox-gl-native-ios/blob/69ddd885126a9650dbe1d20f4b7cee73aa7fc946/platform/macos/scripts/package.sh#L36-L37

The values are automatically calculated based on Git commit history and tags.

## Problem

The build scripts only set `CURRENT_PROJECT_VERSION`, not `DYLIB_CURRENT_VERSION`, which is the setting that matters more for dynamic libraries. (It’s what you see when running `otool -L`.) By default, `DYLIB_CURRENT_VERSION` is supposed to be based on `CURRENT_PROJECT_VERSION`, but we’ve hard-coded it as 1.

`CURRENT_PROJECT_VERSION` and `CURRENT_SHORT_VERSION` only get filled in when building from the command line, not when debugging in Xcode as we normally do.

Moreover, for a time, we also used the scripts to fill in macosapp’s build settings:

https://github.com/mapbox/mapbox-gl-native-ios/blob/69ddd885126a9650dbe1d20f4b7cee73aa7fc946/platform/macos/scripts/package.sh#L60-L61

However, ever since we’ve had to notarize macosapp, I’ve been manually archiving it inside Xcode. Once complete, mapbox/mapbox-gl-native#14754 #204 would automate notarization, allowing us to rely on the script to archive the application again. In the meantime, the version remains at 0.1.0.

version

## Proposed solution

To adopt Apple Generic Versioning, we would set the `CURRENT_PROJECT_VERSION` and `VERSIONING_SYSTEM` build settings as described in [this documentation](https://developer.apple.com/library/archive/qa/qa1827/_index.html).

Then we would add the following step to the [release process](https://github.com/mapbox/mapbox-gl-native/wiki/Releasing-the-Mapbox-iOS-SDK), to be run manually at the same time that we increment the version in the CocoaPods podspecs, so the results would be committed to the repository:

```bash
agvtool bump -all
agvtool new-marketing-version $( git describe --tags --match=ios-v*.*.* --abbrev=0 | sed 's/^ios-v//' )
```

A simple script could run these commands and modify the podspecs.

## Rationale

[Apple Generic Versioning](https://developer.apple.com/library/archive/qa/qa1827/_index.html) is a more-or-less preferred way to manage the build artifacts’ versions. We can continue to calculate the versions the same way, but we’d specify them in a way that applies within Xcode too.

Mapbox’s iOS navigation SDK and Swift service libraries all use Apple Generic Versioning, invoking `agvtool` manually at release time.

/cc @mapbox/maps-ios

1ec5:
Upgraded to [mbgl v1.6.0](https://github.com/mapbox/mapbox-gl-native/releases/tag/maps-v1.6.0). Adopted Apple Generic Versioning on iOS, just like on macOS in #263, so that now both SDKs have the same build numbers and streamlined version updating process.

Updated the Xcode projects and CocoaPods podspecs for Mapbox Maps SDK for iOS v5.9.0-beta.1 and Mapbox Maps SDK for macOS v0.16.0-beta.1.

Fixes #262.

/cc @mapbox/maps-ios


LICENSE.md
```diff
@@ -1,4 +1,4 @@
-mapbox-gl-native-ios Copyright (c) 2014-2020 Mapbox.
+mapbox-gl-native-ios copyright (c) 2014-2020 Mapbox.
```
1ec5, line 1: Really tempted to correct that ASCII `(c)` and `-` into something typographically correct…

1ec5:
`-[MGLResourceTests testOfflineQueryParameterIsAddedForOfflineResource]` and `-[MGLResourceTests testOfflineQueryParameterIsAddedForOfflineResourceForChina]` are failing due to mapbox/mapbox-gl-native#16455.

PR #298, opened 2020-05-06
Make offline database path configurable

### LINKED ISSUE #297: Add global option to specify offline database path
`+[MGLOfflineStorage cacheURLIncludingSubdirectory:]` should respect a new Info.plist key, `MGLOfflineStoragePath`, that specifies the path to cache.db. Most applications shouldn’t need to customize the path, but it may be useful for applications that need to implement sophisticated migration paths between different versions of the application.

/cc @mapbox/maps-ios @zugaldia

1ec5:
Added the `MGLOfflineStorage.databasePath` and `MGLOfflineStorage.databaseURL` properties to obtain the path of the database that contains offline packs and the ambient cache, as well as an `MGLOfflineStorageDatabasePath` Info.plist key to customize this path.

Fixes #297.

/cc @mapbox/maps-ios

### DIFF
--- a/platform/darwin/src/MGLOfflineStorage.h
+++ b/platform/darwin/src/MGLOfflineStorage.h
@@ -201,42 +201,6 @@ MGL_EXPORT
  */
 @property (class, nonatomic, readonly) MGLOfflineStorage *sharedOfflineStorage;
 
-#pragma mark - Adding Contents of File
-
-/**
- Adds the offline packs located at the given file path to offline storage.
- 
- The file must be a valid offline region database bundled with the application
- or downloaded separately.
- 
- The resulting packs are added or updated to the shared offline storage object’s `packs`
- property, then the `completion` block is executed.
- 
- @param filePath A string representation of the file path. The file path must be
- writable as schema updates may be perfomed.
- @param completion The completion handler to call once the contents of the given
- file has been added to offline storage. This handler is executed asynchronously
- on the main queue.
- */
-- (void)addContentsOfFile:(NSString *)filePath withCompletionHandler:(nullable MGLBatchedOfflinePackAdditionCompletionHandler)completion;
-
-/**
- Adds the offline packs located at the given URL to offline storage.
- 
- The file must be a valid offline region database bundled with the application
- or downloaded separately.
- 
- The resulting packs are added or updated to the shared offline storage object’s `packs`
- property, then the `completion` block is executed.
- 
- @param fileURL A file URL specifying the file to add. URL should be a valid system path.
- The file URL must be writable as schema updates may be performed.
- @param completion The completion handler to call once the contents of the given
- file has been added to offline storage. This handler is executed asynchronously
- on the main queue.
- */
-- (void)addContentsOfURL:(NSURL *)fileURL withCompletionHandler:(nullable MGLBatchedOfflinePackAdditionCompletionHandler)completion;
-
 #pragma mark - Accessing the Delegate
 
 /**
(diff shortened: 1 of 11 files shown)


platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -201,54 +201,72 @@ MGL_EXPORT
...
+
+/**
+ The receiver’s delegate.
+
+ An offline storage object sends messages to its delegate to allow it to
+ transform URLs before they are requested from the internet. This can be used
+ add or remove custom parameters, or reroute certain requests to other servers
+ or endpoints.
+ */
+@property(nonatomic, weak, nullable) IBOutlet id<MGLOfflineStorageDelegate> delegate;
```
1ec5, line 214: This isn’t new; I just moved it up above the database merging methods.

platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -153,72 +154,6 @@ - (void)setDelegate:(id<MGLOfflineStorageDelegate>)newValue {
...
- The cache is located in a directory specific to the application, so that packs
- downloaded by other applications don’t count toward this application’s limits.
-
- The cache is located at:
- ~/Library/Application Support/tld.app.bundle.id/.mapbox/cache.db
-
- The subdirectory-less cache was located at:
- ~/Library/Application Support/tld.app.bundle.id/cache.db
- */
-+ (NSURL *)cacheURLIncludingSubdirectory:(BOOL)useSubdirectory {
```
1ec5, line 169: Previously these class methods were sandwiched below certain methods related to `+sharedOfflineStorage` and above `-init`. I’ve consolidated them down where the other database-related methods are implemented and renamed them for consistency, but otherwise I haven’t changed them.

platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -228,26 +163,8 @@ - (instancetype)init {
...
-            NSString *legacyCachePath = [[self class] legacyCachePath];
-            [[NSFileManager defaultManager] moveItemAtPath:legacyCachePath toPath:cachePath error:NULL];
-        }
-
-        // Move the offline file cache from v3.2.x path to a subdirectory that
-        // can be reliably excluded from backups.
-        if (![[NSFileManager defaultManager] fileExistsAtPath:cachePath]) {
-            NSURL *subdirectorylessCacheURL = [[self class] cacheURLIncludingSubdirectory:NO];
-            [[NSFileManager defaultManager] moveItemAtPath:subdirectorylessCacheURL.path toPath:cachePath error:NULL];
-        }
```
1ec5, line 246: These preflight steps have been moved into a lazily initialized `-offlineURL` getter.

platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -299,7 +216,112 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
...
+    if (!_databaseURL) {
+        NSString *customPath = [NSBundle.mainBundle objectForInfoDictionaryKey:MGLOfflineStorageDatabasePathInfoDictionaryKey];
+        if ([customPath isKindOfClass:[NSString class]]) {
+            _databaseURL = [NSURL fileURLWithPath:customPath.stringByStandardizingPath isDirectory:NO relativeToURL:NSBundle.mainBundle.resourceURL];
+            
+            NSURL *directoryURL = _databaseURL.URLByDeletingLastPathComponent;
+            [NSFileManager.defaultManager createDirectoryAtURL:directoryURL
+                                   withIntermediateDirectories:YES
+                                                    attributes:nil
+                                                         error:nil];
```
1ec5, line 235: As with the default code path in `-defaultDatabaseURLIncludingSubdirectory:`, this custom path code path results in the containing directory being created if it doesn’t already exist. However, this code path doesn’t exclude the containing directory from iCloud backups. Info.plist could’ve specified a path in the same folder as other resources. Besides, the developer might decide that the database won’t be so large and that they don’t want to regenerate the database in case the application gets restored from a backup.

platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -631,13 +653,7 @@ - (void)resetDatabaseWithCompletionHandler:(void (^)(NSError *_Nullable error))c
...
 
 - (unsigned long long)countOfBytesCompleted {
-    NSURL *cacheURL = [[self class] cacheURLIncludingSubdirectory:YES];
-    NSString *cachePath = cacheURL.path;
-    if (!cachePath) {
-        return 0;
-    }
-
-    NSDictionary *attributes = [[NSFileManager defaultManager] attributesOfItemAtPath:cachePath error:NULL];
+    NSDictionary *attributes = [[NSFileManager defaultManager] attributesOfItemAtPath:self.databasePath error:NULL];
```
1ec5, line 656: When run from within a unit test bundle before iOS SDK 12.2, `-cacheURLIncludingSubdirectory:` happened to return a stable result only because of a special case in `+[NSBundle mgl_applicationBundleIdentifier]`. Now we can rely on the fact that `databasePath` is memoized.

PR #311, opened 2020-05-26
fix crash on nil completion

gloubibou:
Fences off call to completion() to fix crash on nil completion


captainbarbosa:
Just kicked off CI - I'll follow up with the changelog entry separately.

captainbarbosa:
Tests passed. Thanks for your contribution @gloubibou!

PR #313, opened 2020-06-03
Circleci project setup

Trapster223:
`<changelog>Replace text with suggested changelog entry for this PR or add the 'skip changelog' label.</changelog>`


1ec5:
Hi, this looks like mapbox/mapbox-navigation-ios#2383, but we’re unsure what to do with these PRs. Can you explain what you’re trying to do?

Issue #220, opened 2019-04-03
-[MGLMapView accessibilityElementAtIndex:] performs poorly on Simulator


friedbunny:
When running XCUITests, `-[MGLMapView accessibilityElementAtIndex:]` can occasionally consume 100% of the Simulator’s CPU indefinitely. This can cause the tests to fail if the test harness was waiting for the app to idle before continuing.

Perhaps related to mapbox/mapbox-gl-native#13860 and/or mapbox/mapbox-gl-native-ios#172.

![Screen Shot 2019-04-03 at 1 31 20 PM](https://user-images.githubusercontent.com/1198851/55511264-fc13c700-5614-11e9-83cf-33a8d02ecef9.png)
_[see attached Instruments trace](https://github.com/mapbox/mapbox-gl-native/files/3040629/14324-Instruments.trace.zip)_

### Steps to reproduce

 Run XCUITests on Simulator, e.g., using `ios-sdk-examples` — intermittently pegs CPU.

### Configuration

**Mapbox SDK versions:** `ios-v4.10.0-beta.1`, others
**Device/simulator models:** iPhone SE (iOS 10.3.1), others
**Xcode version:** 10.2

/cc @julianrex @1ec5

julianrex:
Transferred this issue to `mapbox-gl-native-ios` - this issue is still valid. We should look at continuing the work started in https://github.com/mapbox/mapbox-gl-native/pull/15303.

/cc @knov

benjidea:
Hey, is there an update on this?

We (probably) have the same problem, but unfortunately on a real device. 
Once you enable voice over, the CPU is set to 100% and the app crashes when the memory reaches ~2GB.
The app is a bit complex but the ViewController with the MapView doesn't contain any additional elements.

A simple demo project (only one MapView) becomes unusable when Voice Over is activated for ~6 seconds, but does not crash.

julianrex:
This issue was fixed by https://github.com/mapbox/mapbox-gl-native-ios/pull/318

PR #317, opened 2020-06-15
Add query integration test

julianrex:
This is part 1 of 3 PRs that port https://github.com/mapbox/mapbox-gl-native/pull/15303 to this repo. (See also #318, #319.)

This PR adds a query test, but primarily adds a regression test for comparing accessibility elements - to track improvements to `accessibilityElementAtIndex:`. See https://github.com/mapbox/mapbox-gl-native-ios/issues/220 for the original issue.

PR #318, opened 2020-06-15
Refactor accessibility calculations

julianrex:
This is part 2 of 3 PRs that port https://github.com/mapbox/mapbox-gl-native/pull/15303 to this repo. (See also #317, #319.)

This PR, refactors the accessibility element calculations, caching the results. Previously these were needlessly being called continually. This PR addresses the performance concerns. In addition, I have seen fewer crashes.

Fixes https://github.com/mapbox/mapbox-gl-native-ios/issues/220.

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3151,7 +3151,7 @@ public:
     
     // Let the annotation view serve as its own accessibility element.
     MGLAnnotationView *annotationView = annotationContext.annotationView;
-    if (annotationView)//asdf && annotationView.superview)
+    if (annotationView && annotationView.superview)
     {
         return annotationView;
     }
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -2933,12 +2933,14 @@ public:
     
     // Places
     NSArray *sortedPlaceFeatures = nil;
-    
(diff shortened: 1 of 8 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -2889,139 +2861,277 @@ - (CGRect)accessibilityFrame
...
+    // Places
+    NSArray *sortedPlaceFeatures = nil;
+    NSArray *visiblePlaceFeatures;
+    if (!_sortedVisiblePlaceElements) {
+        
+        MGL_SIGNPOST_BEGIN(self.log, signpost, "gae/query-places");
+        NSArray *placeStyleLayerIdentifiers = [self.style.placeStyleLayers valueForKey:@"identifier"];
+        
+        // TODO:
+        /*NSArray **/visiblePlaceFeatures = [self visibleFeaturesInRect:bounds inStyleLayersWithIdentifiers:[NSSet setWithArray:placeStyleLayerIdentifiers]];
```
1ec5, line 2941: What’s the to-do item here?

julianrex: I lifted `visiblePlaceFeatures` above the if, so that I could use it further below.

jmkiley:
@julianrex Made some minor cleanup. I don't see the memory spike with these changes. Will ticket out tail-work around VoiceOver improvements.

PR #319, opened 2020-06-15
Update accessibility calcs

julianrex:
This is part 3 of 3 PRs that port https://github.com/mapbox/mapbox-gl-native/pull/15303 to this repo. (See also #317, #318.)

This PR checks the visibility of the compass to decide whether it should be included in accessibility elements (before it was always included).

It also uses the sorted (by distance from center or user location view) place names in the abbreviated description. This may not be a change that we want to include at this time.


jmkiley:
Most of the changes here took place in #318

PR #336, opened 2020-07-14
Add accessiblePlaceSourceIdentifiers to make layers available to VoiceOver

jmkiley:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/13922

Previously, the source layer identifiers taken into account when generating accessibility elements were hardcoded. This PR:

* Adds an `NSSet` property to `MGLStyle` that allows users to specify additional sources that should be available for VoiceOver.

*  Use `name` attribute value as a fallback value when generating accessibility elements.

Caveats:
 
* This approach works for point features that belong to `MGLVectorStyleLayer`s.

Future work:

* We should allow users to specify which attribute values are available to accessibility.

cc @knov

### DIFF
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -265,11 +265,11 @@ MGL_EXPORT
 
 // TODO: Docs
 /**
- A set containing the custom source layer identifiers for place features available for accessibility. These features must have a `name` attribute.
-
- This set does not include place or road source identifiers included by default.
- */
-@property (nonatomic) NSSet *accessiblePlaceSourceLayerIdentifiers;
+A set containing user-specified source identifiers for point features available for accessibility. The point features must have a `name` attribute and belong to a `MGLVectorStyleLayer`.
+ 
+This set does not include Mapbox source identifiers included by default.
+*/
+@property (nonatomic) NSSet *accessiblePlaceSourceIdentifiers;
 
 /**
  Returns a source with the given identifier in the current style.
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -265,11 +265,11 @@ MGL_EXPORT
 @property (nonatomic, assign) BOOL performsPlacementTransitions;
 
 /**
-A set containing user-specified source identifiers for point features available for accessibility. The point features must have a `name` attribute that matches those specified by <a href="https://www.mapbox.com/vector-tiles/mapbox-streets-v8/#overview">Mapbox Streets</a> source and belong to a `MGLVectorStyleLayer`.
+A set containing user-specified source layer identifiers for point features available for accessibility. The features should have a `MGLVectorTileSource` The point features must have a `name` attribute that matches those specified by <a href="https://www.mapbox.com/vector-tiles/mapbox-streets-v8/#overview">Mapbox Streets</a> source and belong to a `MGLVectorStyleLayer`.
  
-This set does not include Mapbox Streets source identifiers included by default.
+This set does not include Mapbox Streets source identifiers, which are included by default.
 */
-@property (nonatomic) NSSet *accessiblePlaceSourceIdentifiers;
+@property (nonatomic) NSSet <NSString *> *accessiblePlaceSourceLayerIdentifiers;
 
 /**
  Returns a source with the given identifier in the current style.
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -263,11 +263,10 @@ MGL_EXPORT
  */
 @property (nonatomic, assign) BOOL performsPlacementTransitions;
 
-// TODO: Docs
 /**
-A set containing user-specified source identifiers for point features available for accessibility. The point features must have a `name` attribute and belong to a `MGLVectorStyleLayer`.
+A set containing user-specified source identifiers for point features available for accessibility. The point features must have a `name` attribute that matches those specified by <a href="https://www.mapbox.com/vector-tiles/mapbox-streets-v8/#overview">Mapbox Streets</a> source and belong to a `MGLVectorStyleLayer`.
  
-This set does not include Mapbox source identifiers included by default.
+This set does not include Mapbox Streets source identifiers included by default.
 */
 @property (nonatomic) NSSet *accessiblePlaceSourceIdentifiers;
 
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -243,6 +243,7 @@ MGL_EXPORT
  */
 @property (readonly, copy, nullable) NSString *name;
 
+
 #pragma mark Managing Sources
 
 /**
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -265,7 +265,7 @@ MGL_EXPORT
 @property (nonatomic, assign) BOOL performsPlacementTransitions;
 
 /**
-A set containing user-specified source layer identifiers for point features available for accessibility. The features should have a `MGLVectorTileSource` The point features must have a `name` attribute that matches those specified by <a href="https://www.mapbox.com/vector-tiles/mapbox-streets-v8/#overview">Mapbox Streets</a> source and belong to a `MGLVectorStyleLayer`.
+A set containing user-specified source layer identifiers for point features available for accessibility. The features should have a `MGLVectorTileSource` and belong to a source layer. The point features must have a `name` attribute that matches those specified by <a href="https://www.mapbox.com/vector-tiles/mapbox-streets-v8/#overview">Mapbox Streets</a> source and belong to a `MGLVectorStyleLayer`.
  
 This set does not include Mapbox Streets source identifiers, which are included by default.
 */
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -263,6 +263,14 @@ MGL_EXPORT
  */
 @property (nonatomic, assign) BOOL performsPlacementTransitions;
 
+// TODO: Docs
+/**
+ A set containing the custom source layer identifiers for place features available for accessibility. These features must have a `name` attribute.
+
+ This set does not include place or road source identifiers included by default.
+ */
+@property (nonatomic) NSSet *accessiblePlaceSourceLayerIdentifiers;
+
 /**
  Returns a source with the given identifier in the current style.
 
@@ -330,7 +338,6 @@ MGL_EXPORT
  an `NSError` object describing the problem.
  */
 - (BOOL)removeSource:(MGLSource *)source error:(NSError * __nullable * __nullable)outError;
-
 
 #pragma mark Managing Style Layers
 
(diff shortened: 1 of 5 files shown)


platform/darwin/src/MGLStyle.h
```diff
@@ -263,6 +263,13 @@ MGL_EXPORT
  */
 @property (nonatomic, assign) BOOL performsPlacementTransitions;
 
+/**
+A set containing user-specified source identifiers for point features available for accessibility. The point features must have a `name` attribute that matches those specified by <a href="https://www.mapbox.com/vector-tiles/mapbox-streets-v8/#overview">Mapbox Streets</a> source and belong to a `MGLVectorStyleLayer`.
+ 
+This set does not include Mapbox Streets source identifiers included by default.
+*/
+@property (nonatomic) NSSet *accessiblePlaceSourceIdentifiers;
```
1ec5, line 272: The type should be qualified with the type of each element: `NSSet <NSString *> *`.

1ec5, line 272: Rename the property to `accessiblePlaceSourceLayerIdentifiers` to reflect the fact that it contains source _layer_ identifiers, not source identifiers.

It’s also worth pointing out that this property only applies to MGLVectorTileSource objects or vector tile sources in style JSON. It isn’t possible for an MGLShapeSource, MGLComputedShapeSource, or GeoJSON source in style JSON to have source layers. If we want shape sources to participate in this feature, we’d have to add a second property that is a set of MGLShapeSource and MGLComputedShapeSource objects to also include.

platform/ios/src/MGLMapAccessibilityElement.mm
```diff
@@ -52,6 +52,9 @@ - (instancetype)initWithAccessibilityContainer:(id)container feature:(id<MGLFeat
         NSString *nameAttribute = [NSString stringWithFormat:@"name_%@", languageCode];
         NSString *name = [feature attributeForKey:nameAttribute];
 
+        if (name == nil && [feature attributeForKey:@"name"] != nil) {
+            name = [feature attributeForKey:@"name"];
+        }
```
1ec5, line 57: This fallback should probably go further down, after the transliteration step. The transliteration step depends on knowing the language the name is in, but we don’t know the language of the language-agnostic `name` property.

1ec5: On second thought, Streets source v8 comes with a [`name_script`](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#name_script-text) property that we can feed into `-mgl_stringByTransliteratingIntoScript:` in the event that we don’t know the language. So call it with the language code for `name_*` but with `name_script` for `name`.

platform/ios/src/MGLMapView.mm
```diff
@@ -3083,8 +3083,16 @@ - (void)generateAccessibilityElementsIfNeeded
...
+
+              NSString *languageCode = [MGLVectorTileSource preferredMapboxStreetsLanguage];
+              NSString *nameAttribute = [NSString stringWithFormat:@"name_%@", languageCode];
+              NSString *name = [placeFeature attributeForKey:nameAttribute];
+
+              if (name == nil && [placeFeature attributeForKey:@"name"] != nil) {
+                  name = [placeFeature attributeForKey:@"name"];
+              }
+
+                if (![placesSet containsObject:name] && name != nil) {
```
1ec5, line 3095: Nit: A set can’t contain `nil`, so the second part of this check is redundant.

platform/darwin/src/MGLStyle.mm
```diff
@@ -625,12 +627,26 @@ - (void)localizeLabelsIntoLocale:(nullable NSLocale *)locale {
...
 
+- (void)setAccessiblePlaceSourceIdentifiers:(NSSet *)accessiblePlaceSourceLayerIdentifiers {
+    if (![_accessiblePlaceSourceIdentifiers isEqualToSet:accessiblePlaceSourceLayerIdentifiers]) {
+        _accessiblePlaceSourceIdentifiers = accessiblePlaceSourceLayerIdentifiers;
+    }
+}
+
+- (NSSet *)accessiblePlaceSourceIdentifiers {
+    return _accessiblePlaceSourceIdentifiers;
+}
```
1ec5, line 638: I’m not seeing anything out of the ordinary about this getter and setter, so I think it should be OK to leave out this custom implementation.

platform/darwin/src/MGLStyle.mm
```diff
@@ -122,6 +122,8 @@ + (NSURL *)name##StyleURL##WithVersion:(NSInteger)version { \
 static_assert(6 == mbgl::util::default_styles::numOrderedStyles,
               "mbgl::util::default_styles::orderedStyles and MGLStyle have different numbers of styles.");
 
+@synthesize accessiblePlaceSourceIdentifiers = _accessiblePlaceSourceIdentifiers;
```
1ec5, line 125: It shouldn’t be necessary to synthesize this property explicitly.

platform/darwin/src/MGLStyle.mm
```diff
@@ -625,12 +627,26 @@ - (void)localizeLabelsIntoLocale:(nullable NSLocale *)locale {
...
-    
+
     NSSet *placeSourceLayerIdentifiers = [NSSet setWithObjects:@"marine_label", @"country_label", @"state_label", @"place_label", @"water_label", @"poi_label", @"rail_station_label", @"mountain_peak_label", @"natural_label", @"transit_stop_label", nil];
+    if (self.accessiblePlaceSourceIdentifiers == nil) {
+        _accessiblePlaceSourceIdentifiers = [NSMutableSet set];
+    }
+
     NSPredicate *isPlacePredicate = [NSPredicate predicateWithBlock:^BOOL (MGLVectorStyleLayer * _Nullable layer, NSDictionary<NSString *, id> * _Nullable bindings) {
-        return [layer isKindOfClass:[MGLVectorStyleLayer class]] && [streetsSourceIdentifiers containsObject:layer.sourceIdentifier] && [placeSourceLayerIdentifiers containsObject:layer.sourceLayerIdentifier];
+        return [layer isKindOfClass:[MGLVectorStyleLayer class]] && (([streetsSourceIdentifiers containsObject:layer.sourceIdentifier] && [placeSourceLayerIdentifiers containsObject:layer.sourceLayerIdentifier]) || [self.accessiblePlaceSourceIdentifiers containsObject:layer.sourceIdentifier]);
```
1ec5, line 649: This method returns an array of identifiers of layers within sources used by layers, not an array of identifiers of sources used by layers. For example, it returns `country_label`, not `composite`.

platform/darwin/src/MGLStyle.mm
jmkiley: Thanks, I think that's leftover from when I was taking a different approach! Will clean it up.

platform/ios/src/MGLMapAccessibilityElement.mm
jmkiley: It doesn't look like the values for `name_script` match the values `-mgl_stringByTransliteratingIntoScript:` expects. For example, `name_script` returns Latin, while `-mgl_stringByTransliteratingIntoScript:` expects `Latn`. I'll move the check for `name` a little further down.

jmkiley:
@1ec5 thank you so much for your review! I think I have addressed everything.

platform/ios/src/MGLMapAccessibilityElement.mm
1ec5: `-[NSString stringByApplyingTransform:]` expects an [ICU transliterator identifier](http://userguide.icu-project.org/transforms/general#TOC-Transliterator-Identifiers). `Latin` is a valid identifier – probably most of those `name_script` values are valid identifiers, so it would be fine to dump it into the method and see what comes out.

platform/darwin/src/MGLStyle.mm
```diff
@@ -627,10 +627,14 @@ - (void)localizeLabelsIntoLocale:(nullable NSLocale *)locale {
...
-    
+
     NSSet *placeSourceLayerIdentifiers = [NSSet setWithObjects:@"marine_label", @"country_label", @"state_label", @"place_label", @"water_label", @"poi_label", @"rail_station_label", @"mountain_peak_label", @"natural_label", @"transit_stop_label", nil];
+    if (self.accessiblePlaceSourceLayerIdentifiers == nil) {
+        _accessiblePlaceSourceLayerIdentifiers = [NSMutableSet set];
+    }
+
     NSPredicate *isPlacePredicate = [NSPredicate predicateWithBlock:^BOOL (MGLVectorStyleLayer * _Nullable layer, NSDictionary<NSString *, id> * _Nullable bindings) {
-        return [layer isKindOfClass:[MGLVectorStyleLayer class]] && [streetsSourceIdentifiers containsObject:layer.sourceIdentifier] && [placeSourceLayerIdentifiers containsObject:layer.sourceLayerIdentifier];
+        return [layer isKindOfClass:[MGLVectorStyleLayer class]] && (([streetsSourceIdentifiers containsObject:layer.sourceIdentifier] && [placeSourceLayerIdentifiers containsObject:layer.sourceLayerIdentifier]) || [self.accessiblePlaceSourceLayerIdentifiers containsObject:layer.sourceLayerIdentifier]);
```
1ec5, line 637: Should we limit this additional check to only non-Streets sources? That would avoid accidentally enabling the feature for Streets layers that happen to have name collisions; on the other hand, it would prevent developers from potentially enabling the feature for more Streets layers we didn’t expect to be labeled but that the style happens to label.

platform/ios/src/MGLMapAccessibilityElement.mm
```diff
@@ -52,6 +52,11 @@ - (instancetype)initWithAccessibilityContainer:(id)container feature:(id<MGLFeat
         NSString *nameAttribute = [NSString stringWithFormat:@"name_%@", languageCode];
         NSString *name = [feature attributeForKey:nameAttribute];
 
+        if (name == nil && [feature attributeForKey:@"name"] != nil) {
+            name = [feature attributeForKey:@"name"];
+            languageCode = [feature attributeForKey:@"name_script"];
```
1ec5, line 57: This would pass `name_script` into `+[NSOrthography mgl_dominantScriptForMapboxStreetsLanguage:]`, which expects a language code to turn into a script code. But `name_script` is already a script code, so I think we need to pass it directly into `-[NSString(MGLAdditions) mgl_stringByTransliteratingIntoScript:]`.

jmkiley: Thanks! Updated to set `dominantScript` rather than `languageCode`.

platform/ios/src/MGLMapAccessibilityElement.mm
```diff
@@ -52,16 +52,17 @@ - (instancetype)initWithAccessibilityContainer:(id)container feature:(id<MGLFeat
...
-            languageCode = [feature attributeForKey:@"name_script"];
+            dominantScript = [feature attributeForKey:@"name_script"];
         }
 
         // If a feature hasn’t been translated into the preferred language, it
         // may be in the local language, which may be written in another script.
         // Attempt to transform to the script of the preferred language, keeping
         // the original string if no transform exists or if transformation fails.
-        NSString *dominantScript = [NSOrthography mgl_dominantScriptForMapboxStreetsLanguage:languageCode];
+        dominantScript = [NSOrthography mgl_dominantScriptForMapboxStreetsLanguage:languageCode];
```
1ec5, line 65: This avoids overriding the line above that sets `dominantScript` based on `name_script`:

```suggestion
        if (!dominantScript) {
            dominantScript = [NSOrthography mgl_dominantScriptForMapboxStreetsLanguage:languageCode];
        }
```

PR #349, opened 2020-07-24
[ios] Add Xcode 11.6.0, 12.0 to nightly jobs.

julianrex:
This PR updates Xcode versions for build/testing:

For regular commits, CircleCI now uses Xcode `11.3.1` and `11.5.0`. (These being the two latest stable Xcode versions on Catalina and Mojave.)

For nightly builds, it uses Xcode `11.3.1`, `11.4.1`, `11.5.0`, `11.6.0`, `12.0.0`. i.e. including recent betas.

It now also builds for Release (as well as Debug) - though some tests have been modified since asserts are compiled out in Release configuration.

This PR does not update any of the sanitize jobs, nor does it change the version of Xcode that releases are built with.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,8 +19,7 @@ workflows:
       - ios-build:
           matrix:
             parameters:
-              #xcode: ["11.3.1", "11.5.0"]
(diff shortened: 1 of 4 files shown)


julianrex:
cc @nishant-karajgikar

PR #358, opened 2020-07-30
Adds platform observable interface

julianrex:
This PR adds:
-  `MGLObservable` protocol that declares methods that conforming classes should implement if they want to observe low-level `MGLEvent`s. Currently only `MGLMapView` implements `MGLObservable`.
- `MGLObserver` base class that developers should subclass and use with the above's subscribe and unsubscribe methods. An `MGLObserver` will receive a generic `MGLEvent`.

```
<changelog>Added `MGLObserver`, `MGLObservable` and `MGLEvent` to monitor events from the map view.</changelog>
```


platform/darwin/src/MGLObserver_Private.h
```diff
@@ -0,0 +1,40 @@
...
+public:
+    Observer(MGLObserver *observer_): observer(observer_) {}
+    virtual ~Observer() = default;
+    virtual void notify(const ObservableEvent& event);
+    virtual std::size_t id() const;
+
+protected:
+
+    /// Cocoa observer that this adapter bridges to.
+    __weak MGLObserver *observer = nullptr;
```
alexshalamov, line 22: if `observer` is GCd, who will unsubscribe `mbgl::Observer` from the map?

platform/darwin/src/MGLEvent.mm
```diff
@@ -0,0 +1,107 @@
...
+                                                                                 (event.end - steadyClockNow));
+    
+    auto beginTime  = std::chrono::duration<double, std::ratio<1>>(begin.time_since_epoch()).count();
+    auto endTime    = std::chrono::duration<double, std::ratio<1>>(end.time_since_epoch()).count();
+
+    _type   = (MGLEventType)[NSString stringWithUTF8String:event.type.c_str()];
+    _begin  = beginTime;
+    _end    = endTime;
+//    _begin = [NSDate dateWithTimeIntervalSince1970:beginTime];
+//    _end   = [NSDate dateWithTimeIntervalSince1970:endTime];
```
alexshalamov, line 65: can be removed?

platform/darwin/src/MGLEvent.mm
```diff
@@ -0,0 +1,107 @@
+#import "MGLEvent_Private.h"
+
+const MGLEventType MGLEventTypeResourceRequest = @"resource-request";
+
+#pragma mark - Value conversion
+
+// From value.md
+static id MGLJSONObjectFromMapboxBaseValue(const mapbox::base::Value &value) {
+    // Supported types are `int`, `uint`, `bool`, `double`, `array`, `object` and `string`.
+    return value.match(
```
alexshalamov, line 10: I'm wondering if there is another value->id converter in the repository, maybe in expression or style code?

platform/darwin/src/MGLObserver_Private.h
julianrex: This is handled by documentation of `MGLObservable` - the expectation is that subscribe will retain the `MGLObserver` (and unsubscribe will release). This is done by storing the observer in a set.

The assert in the `MGLObserver` dealloc is for development, to help identify cases where we have neglected to retain the observers.

Another option that we might consider is to extend this, so that `MGLObserver` isn't actually public - and instead we use a notification handler. This would allow us to control `MGLObserver` completely, and would feel a little more idiomatic. (In fact, as you've noticed the base class already has a handler - this may or may not make it through to release).

platform/darwin/src/MGLEvent.mm
julianrex: Will remove in a later PR. It's there as a reminder to see which feels more appropriate. I'm not sure that creating NSDates (which can be expensive) is worth it in this instance.

platform/darwin/src/MGLEvent.mm
julianrex: You're right, I think we do. I was looking at `mbgl::Value` and assuming this was different from `mapbox::base::Value`.

julianrex: @alexshalamov do you think we can reuse [this one](https://github.com/mapbox/mapbox-gl-native-ios/blob/0f6c746a447ff3da15559f8cfd6c4d045756b9d0/platform/darwin/src/MGLStyleValue.mm#L8)? Looks like we'd need to add support for `uint64_t`.

julianrex: Done.

platform/darwin/src/MGLEvent.mm
julianrex: Done.

platform/ios/src/Mapbox.h
```diff
@@ -7,70 +7,72 @@ FOUNDATION_EXPORT MGL_EXPORT double MapboxVersionNumber;
 /// Project version string for Mapbox.
 FOUNDATION_EXPORT MGL_EXPORT const unsigned char MapboxVersionString[];
 
-#import "MGLAnnotationView.h"
 #import "MGLAccountManager.h"
 #import "MGLAnnotation.h"
 #import "MGLAnnotationImage.h"
+#import "MGLAnnotationView.h"
+#import "MGLAttributedExpression.h"
```
jmkiley, line 14: Is this just alphabetizing the headers?

julianrex: yup

PR #361, opened 2020-08-03
iOS 14 location privacy settings update

### LINKED ISSUE #337: Update MGLLocationManager to iOS 14
MGLLocationManager has to update the location authorization API to reflect [iOS 14](https://developer.apple.com/documentation/corelocation/cllocationmanager/3600215-accuracyauthorization?changes=latest_minor) changes. 

The main changes are:
- [x] Add `accuracyAuthorization` property.
- [x] Add the equivalent of `func requestTemporaryFullAccuracyAuthorization(withPurposeKey: String)` to `MGLLocationManager`
- [x] Change from class to instance the method `authorizationStatus()` 
~~- [ ] Modify `MGLMapView` to request the accuracy authorization~~
- [x] Update `MGLLocationManager` documentation to account the state management involving `accuracyAuthorization`

fabian-guerra:
Fixes #337 

iOS 14 has a new privacy setting related to location accuracy. It means that users may be able to revoke the level of location accuracy the app has permission to use. When this happens apps will not be able to change manually this setting. However in order to request full location accuracy for a short period of time this pr introduces `requestTemporaryFullAccuracyAuthorizationWithPurposeKey` which is on pair of `CLLocationManager. requestTemporaryFullAccuracyAuthorizationWithPurposeKey`. Developers can check the location accuracy `MGLLocationManager. accuracyAuthorization` and if it is `CLAccuracyAuthorizationReducedAccuracy` then they can call `requestTemporaryFullAccuracyAuthorizationWithPurposeKey` to gain full accuracy for a short period of time. Enough to complete their use case.

### DIFF
--- a/platform/darwin/src/MGLLocationManager.m
+++ b/platform/darwin/src/MGLLocationManager.m
@@ -38,7 +38,11 @@
 }
 
 - (CLAuthorizationStatus)authorizationStatus {
-    return [CLLocationManager authorizationStatus];
+    if (@available(iOS 14.0, *)) {
+        return self.locationManager.authorizationStatus;
+    } else {
+        return [CLLocationManager authorizationStatus];
+    }
 }
 
 - (void)setActivityType:(CLActivityType)activityType {
@@ -47,6 +51,14 @@
 
 - (CLActivityType)activityType {
     return self.locationManager.activityType;
+}
+
+- (CLAccuracyAuthorization)accuracyAuthorization {
+    if (@available(iOS 14.0, *)) {
+        return self.locationManager.accuracyAuthorization;
+    } else {
+        return CLAccuracyAuthorizationFullAccuracy;
+    }
 }
 
 - (void)dismissHeadingCalibrationDisplay {
--- a/platform/darwin/src/MGLLocationManager.m
+++ b/platform/darwin/src/MGLLocationManager.m
@@ -61,6 +61,12 @@
     }
 }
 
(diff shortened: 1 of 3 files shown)


platform/darwin/src/MGLLocationManager.m
```diff
@@ -49,6 +57,22 @@ - (CLActivityType)activityType {
...
+    if (@available(iOS 14.0, *)) {
+        return self.locationManager.accuracyAuthorization;
+    } else {
+        return CLAccuracyAuthorizationFullAccuracy;
+    }
+}
+
+- (void)requestTemporaryFullAccuracyAuthorizationWithPurposeKey:(NSString *)purposeKey {
+    if (@available(iOS 14.0, *)) {
+        [self.locationManager requestTemporaryFullAccuracyAuthorizationWithPurposeKey:purposeKey];
```
nishant-karajgikar, line 71: I'm guessing we need to add a new purpose key in the plist now?

nishant-karajgikar: DO we need to make any changes to documentation to reflect this?

captainbarbosa:
I think this requires a changelog entry.

platform/darwin/src/MGLLocationManager.m
fabian-guerra: The key comes from Apple. But yes we will have to modify our docs.

Issue #370, opened 2020-08-07
Crash in logging (Debug and Error levels)


julianrex:
Variable arguments are not being passed to `debugLog` and `errorLog` here:

https://github.com/mapbox/mapbox-gl-native-ios/blob/709ea4cebc9aac49dd1d1c317b1a2221fed9c8e7/platform/darwin/src/MGLNetworkConfiguration.m#L130-L136

This will result in a crash if the logging level is set to include debug or errors.

We need to add a method to `MGLLoggingConfiguration`, along the lines of:

```objc
- (void)logCallingFunction:(const char *)callingFunction functionLine:(NSUInteger)functionLine messageType:(MGLLoggingLevel)type format:(id)messageFormat arguments:(va_list)argList;
```

(and associated calling macros).

### Configuration

**Mapbox SDK versions:** 6.1.0-beta.1 (though was introduced in https://github.com/mapbox/mapbox-gl-native-ios/pull/228)

PR #373, opened 2020-08-13
Add Logging level tests (and vararg version of logCallingFunction)

### LINKED ISSUE #347: MGLNetworkConfiguration#errorLog drops error details
Nearly a PR, but I don't have time to get the build running locally at the moment.

> We've been seeing errors logged from `[MGLNetworkConfiguration errorLog:]` with incomplete messages that look like `Requesting: (null) failed with error: <MGLNetworkConfiguration: 0x60000291ea60>`.
> 
> Why is the URL `(null)`? Why is the error an instance of `MGLNetworkConfiguration`?

This is a companion issue to https://github.com/mapbox/mapbox-gl-native/issues/16511, which covers the first part of the problem. Until that's fixed, fixing this won't do anything because the details will be dropped upstream.

The implementation of MGLNetworkConfiguration takes a format string and variadic arguments, but it doesn't pass the variadic args along when calling `MGLLogError`:

https://github.com/mapbox/mapbox-gl-native-ios/blob/171ff9484a48c39151a4bb793911889996f54156/platform/darwin/src/MGLNetworkConfiguration.m#L134-L136

julianrex:
This PR adds unit tests as a part of: https://github.com/mapbox/mapbox-gl-native-ios/issues/370

See gl-native for the associated fix for #370 (which uses `Log::Observer` instead of calling down to `-[MGLNetworkConfiguration debugLog:]` etc).

Fixes #347 (along with next gl-native release)
Fixes https://github.com/mapbox/mapbox-gl-native/issues/16511

### DIFF
--- a/platform/darwin/src/MGLNetworkConfiguration.m
+++ b/platform/darwin/src/MGLNetworkConfiguration.m
@@ -127,13 +127,15 @@ NSString * const kMGLDownloadPerformanceEvent = @"mobile.performance_trace";
     [self sendEventForURLResponse:response withAction:@"cancel"];
 }
 
-- (void)debugLog:(NSString *)format, ... {
-    MGLLogDebug(format);
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-implementations"
+// TODO: Remove in future release.
+- (void)debugLog:(nonnull NSString *)format, ... {
 }
 
-- (void)errorLog:(NSString *)format, ... {
-    MGLLogError(format);
+- (void)errorLog:(nonnull NSString *)format, ... {
 }
+#pragma clang diagnostic pop
 
 #pragma mark - Event management
 
--- a/platform/ios/test/MBXLoggingConfigurationTests.m
+++ b/platform/ios/test/MBXLoggingConfigurationTests.m
@@ -1,0 +1,155 @@
+#import <XCTest/XCTest.h>
+@import Mapbox;
+#import "MGLLoggingConfiguration_Private.h"
+
+@interface MBXLoggingConfigurationTests : XCTestCase
+@property (nonatomic, assign) MGLLoggingLevel loggingLevel;
+@end
+
+@implementation MBXLoggingConfigurationTests
+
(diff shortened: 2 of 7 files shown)


platform/darwin/src/MGLNetworkConfiguration.m
```diff
@@ -127,13 +127,15 @@ - (void)cancelDownloadEventForResponse:(NSURLResponse *)response {
     [self sendEventForURLResponse:response withAction:@"cancel"];
 }
 
-- (void)debugLog:(NSString *)format, ... {
-    MGLLogDebug(format);
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-implementations"
+// TODO: Remove in future release.
```
fabian-guerra, line 132: Why are we removing this?

platform/ios/test/MBXLoggingConfigurationTests.m
```diff
@@ -0,0 +1,165 @@
...
+        @(MGLLoggingLevelWarning),
+        @(MGLLoggingLevelError),
+        @(MGLLoggingLevelFault)
+    ]];
+}
+
+- (void)testLoggingLevelInfo {
+    [self internalTestLoggingLevel:MGLLoggingLevelInfo
+             expectingLogForLevels:@[
+//      @(MGLLoggingLevelDebug),
```
fabian-guerra, line 75: nit: please remove commented code.

platform/darwin/src/MGLNetworkConfiguration.m
julianrex: To bypass a catch-22; I marked the protocol methods as deprecated, and so the compiler warns (errors) if you add them. But if they're missing then the compiler errors about an incomplete conformance. 🤕

platform/ios/test/MBXLoggingConfigurationTests.m
julianrex: Tempted to leave for the moment, as in this case it's kinda documentation when seen alongside the other tests.

PR #375, opened 2020-08-14
Re-enable bitcode for the release branch

jmkiley:
This PR:

    Re-enables bitcode for the dynamic target.
    Removes the linker flags added in #368 

cc @julianrex


julianrex:
@nishant-karajgikar needed for beta.2

PR #376, opened 2020-08-17
Location changes delegate methods.

fabian-guerra:
Adds delegate methods to let developers know about location accuracy changes.

### DIFF
--- a/platform/darwin/src/MGLLocationManager.m
+++ b/platform/darwin/src/MGLLocationManager.m
@@ -136,5 +136,9 @@
         [self.delegate locationManager:self didFailWithError:error];
     }
 }
-
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 140000
+- (void)locationManagerDidChangeAuthorization:(CLLocationManager *)manager {
+    [self.delegate locationManagerDidChangeAuthorization:self];
+}
+#endif
 @end
--- a/platform/ios/app/MBXOrnamentsViewController.m
+++ b/platform/ios/app/MBXOrnamentsViewController.m
@@ -2,7 +2,7 @@
 
 #import "MBXOrnamentsViewController.h"
 
-@interface MBXOrnamentsViewController ()<MGLMapViewDelegate>
+@interface MBXOrnamentsViewController ()
 
 @property (nonatomic) MGLMapView *mapView;
 @property (nonatomic) NSTimer *timer;
@@ -64,7 +64,7 @@
                        zoomLevel:16
                        direction:30
                         animated:NO];
-    mapView.delegate = self;
+    
     mapView.showsScale = YES;
     [self.view addSubview:mapView];
 
--- a/platform/ios/src/MGLMapViewDelegate.h
+++ b/platform/ios/src/MGLMapViewDelegate.h
@@ -368,7 +368,6 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (CGPoint)mapViewUserLocationAnchorPoint:(MGLMapView *)mapView;
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 140000
 /**
  Tells the delegate that the map's location updates accuracy authorization has changed.
  
@@ -376,11 +375,10 @@ NS_ASSUME_NONNULL_BEGIN
  requesting location permissions or in privacy settings.
  
  @param mapView The map view that changed its location accuracy authorization.
- @param accuracyAuthorization The new accuracy authorization value.
+ @param manager The location manager reporting the update.
  
  */
-- (void)mapView:(MGLMapView *)mapView didChangeAccuracyAuthorization:(CLAccuracyAuthorization)accuracyAuthorization;
-#endif
+- (void)mapView:(MGLMapView *)mapView didChangeLocationManagerAuthorization:(id<MGLLocationManager>)manager API_AVAILABLE(ios(14));
 
 #pragma mark Managing the Appearance of Annotations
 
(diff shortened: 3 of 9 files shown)


platform/darwin/src/MGLLocationManager.m
```diff
@@ -38,7 +38,15 @@ - (CLLocationAccuracy)desiredAccuracy {
...
 - (CLAuthorizationStatus)authorizationStatus {
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 140000
+    if (@available(iOS 14.0, *)) {
+        return self.locationManager.authorizationStatus;
+    } else {
+        return kCLAuthorizationStatusNotDetermined;
+    }
+#else
     return [CLLocationManager authorizationStatus];
+#endif
```
neelmistry94, line 49: I'm not sure if this stylistic or a necessity, but do we need the pre processor #if statements?

platform/darwin/src/MGLLocationManager.m
```diff
@@ -38,7 +38,15 @@ - (CLLocationAccuracy)desiredAccuracy {
...
 }
 
 - (void)setActivityType:(CLActivityType)activityType {
```
neelmistry94, line 52: I know you didn't change this, but the function says set activity type, but you are just returning an existing activity type, that doesn't make sense

platform/darwin/src/MGLLocationManager.m
```diff
@@ -49,6 +57,22 @@ - (CLActivityType)activityType {
...
+    if (@available(iOS 14.0, *)) {
+        return self.locationManager.accuracyAuthorization;
+    } else {
+        return CLAccuracyAuthorizationFullAccuracy;
+    }
+}
+
+- (void)requestTemporaryFullAccuracyAuthorizationWithPurposeKey:(NSString *)purposeKey {
+    if (@available(iOS 14.0, *)) {
+        [self.locationManager requestTemporaryFullAccuracyAuthorizationWithPurposeKey:purposeKey];
```
neelmistry94, line 71: wouldn't this be an endless loop/recursive call if you are on iOS14. 
This looks like the implementation is calling the method itself over and over again, and if you are on iOS 14, that will happen non stop

platform/darwin/src/MGLLocationManager.m
```diff
@@ -112,5 +136,9 @@ - (void)locationManager:(CLLocationManager *)manager didFailWithError:(NSError *
         [self.delegate locationManager:self didFailWithError:error];
     }
 }
-
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 140000
+- (void)locationManagerDidChangeAuthorization:(CLLocationManager *)manager {
+    [self.delegate locationManagerDidChangeAuthorization:self];
```
neelmistry94, line 141: same comment as above

platform/darwin/src/MGLLocationManager.m
fabian-guerra: Pragma marks are there to make it compatible with Xcode 11.

platform/darwin/src/MGLLocationManager.m
fabian-guerra: The default implementation of `MGLLocationManager` is a wrapper of `CLLocationManager.

platform/darwin/src/MGLLocationManager.m
fabian-guerra: Same comment as above.

platform/darwin/src/MGLLocationManager.m
fabian-guerra: Same as https://github.com/mapbox/mapbox-gl-native-ios/pull/376#discussion_r472232475

platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -17,7 +17,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+ 
+ This method is called after the user changes location accuracy authorization when
+ requesting location permissions or in privacy settings.
+ 
+ @param mapView The map view that changed its location accuracy authorization.
+ @param accuracyAuthorization The new accuracy authorization value.
+ 
+ */
+- (void)mapView:(MGLMapView *)mapView didChangeAccuracyAuthorization:(CLAccuracyAuthorization)accuracyAuthorization;
+#endif
```
nishant-karajgikar, line 32: We shouldn't need the xcode 11 compatibility test here as this is just a wrapper.

Also -- shouldn't this method be optional? There can be consumers of the map who don't care about user location entirely.

platform/ios/app/MBXOrnamentsViewController.m
```diff
@@ -64,7 +64,7 @@ - (void)viewDidLoad {
                        zoomLevel:16
                        direction:30
                         animated:NO];
-    mapView.delegate = self;
+    
     mapView.showsScale = YES;
```
nishant-karajgikar, line 68: Why are we removing this?

platform/ios/src/MGLMapViewDelegate.h
fabian-guerra: > We shouldn't need the xcode 11 compatibility test here as this is just a wrapper.

In order to build the SDK with either Xcode 11/12 this has to be here. Unless we drop Xcode 11 support.

> Also -- shouldn't this method be optional? There can be consumers of the map who don't care about user location entirely.

Right. However with the privacy changes this has to be required to let developers be aware that they will have to handle that, even tho they don't use location updates.

platform/ios/app/MBXOrnamentsViewController.m
fabian-guerra: This was dead code in the examples that surfaced with the introduction of the new delegate method.

nishant-karajgikar:
Thanks for making the change, @fabian-guerra . I won't be able to test this on an iOS14 device just yet (running into trouble downloading the latest xcode beta). Feel free to merge this and we can address anything we see during test in a subsequent PR.

PR #381, opened 2020-08-19
Add an approximate location view.

fabian-guerra:
Introduces a new approximate location view when `MGLLocationManager.accuracyAuthorization` is set to `CLAccuracyAuthorizationReducedAccuracy`.

approximatelocatiion


platform/ios/src/MGLFaux3DUserLocationAnnotationView.m
```diff
@@ -433,6 +464,74 @@ - (void)drawDot
...
+        _puckArrow = nil;
+        
+        _approximateModeActivated = YES;
+    }
+
+    // update accuracy ring (if zoom or horizontal accuracy have changed)
+    if (_approximateLayer && (_oldZoom != self.mapView.zoomLevel || _oldHorizontalAccuracy != self.userLocation.location.horizontalAccuracy))
+    {
+        CGFloat borderWidth = 2;
+        if (self.mapView.zoomLevel < 7) {
```
neelmistry94, line 492: can we move this zoom level value into a const in case we have to change in the future
maybe like `const int maxApproximateZoomLevel = 7` <-- or w.e the correct obj-c syntax is

platform/ios/src/MGLFaux3DUserLocationAnnotationView.m
```diff
@@ -49,10 +55,33 @@ - (void)update
 
     if (CLLocationCoordinate2DIsValid(self.userLocation.coordinate))
     {
-        (self.mapView.userTrackingMode == MGLUserTrackingModeFollowWithCourse) ? [self drawPuck] : [self drawDot];
-        [self updatePitch];
+        if (@available(iOS 14, *)) {
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 140000
+            if (self.mapView.locationManager.accuracyAuthorization == CLAccuracyAuthorizationFullAccuracy) {
```
bamx23, line 60: `accuracyAuthorization` is optional and PassiveLocationManager in Navigation SDK hasn't implemented this property in 1.0 release. This leads to crashes with Nav SDK 1.0 and Maps SDK 6.2.0 on iOS 14.

1ec5: mapbox/mapbox-navigation-ios#2554

1ec5: mapbox/mapbox-gl-native-ios#470

PR #391, opened 2020-08-24
Fix typo in macOS map SDK installation instructions

1ec5:
The macOS map SDK documentation and Carthage binary project specification remained on the GitHub Pages site corresponding to the mapbox/mapbox-gl-native repository even after the codebase moved to mapbox/mapbox-gl-native-ios. An overzealous find-and-replace caused the installation instructions to be incorrect, preventing anyone from installing the SDK using Carthage.

Separately, I’ll go and manually patch the gh-pages branch to point to the right place.

/cc @mapbox/maps-ios @ZiZasaurus

### DIFF
--- a/platform/macos/docs/pod-README.md
+++ b/platform/macos/docs/pod-README.md
@@ -27,7 +27,7 @@ There are three ways to install the Mapbox Maps SDK for macOS:
 The Mapbox Maps SDK for macOS is a binary-only dependency, so you’ll need Carthage 0.19 or above. In your [Cartfile](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#binary-only-frameworks), specify this dependency (plus an optional version requirement):
 
 ```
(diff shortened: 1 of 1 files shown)


1ec5:
debafe14d78dd31a2a4a0521569c132154f1b0f4 patches the docsets for v0.15.0-alpha.1 through v0.16.0, inclusive.

1ec5:
Even after this fix, the URLs in that binary project specification were pointing to the mapbox-gl-native repository, instead of the mapbox-gl-native-ios repository where the binaries had been uploaded. I just fixed this oversight in 6ba72e946c9ec4943c198613928e8feb9f606b8e.

If you’ve been running into this issue, you should be able to install the SDK using `carthage update --platform macOS` as usual. However, if Carthage is still picking up the old, nonexistent URL and failing to download it, clear your Carthage caches (plural) using:

```bash
rm -rf ~/Library/Caches/carthage ~/Library/Caches/org.carthage.CarthageKit/binaries/Mapbox*
```

Issue #385, opened 2020-08-20
Explore: Expose styling for default user location view


fabian-guerra:
With the introduction of a new user location state in the approximate location [pr](https://github.com/mapbox/mapbox-gl-native-ios/pull/381) it may be necessary to expose the default puck/arrow/approximate location to style updates such as border width, color, opacity.

- [ ] Evaluate the impact
- [ ] Review alternatives (Delegate, Protocol, etc)

PR #403, opened 2020-08-31
Expose styling options for the user location annotation.

fabian-guerra:
Fixes https://github.com/mapbox/mapbox-gl-native-ios/issues/385

Styling example:
### Halo
puck_halo
### Arrow
arrow
### Puck
puck
### Approximate
aproximate

### DIFF
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -2368,8 +2368,8 @@ CLLocationCoordinate2D randomWorldCoordinate() {
     }
 }
 
-- (MGLUserLocationAnnotationViewStyle)mapViewStyleForDefaultUserLocationAnnotationView:(MGLMapView *)mapView {
-    MGLUserLocationAnnotationViewStyle style;
+- (MGLUserLocationAnnotationViewStyle *)mapViewStyleForDefaultUserLocationAnnotationView:(MGLMapView *)mapView {
+    MGLUserLocationAnnotationViewStyle *style = [[MGLUserLocationAnnotationViewStyle alloc] init];
     
     style.approximateHaloFillColor = UIColor.redColor;
     style.approximateHaloOpacity = 0.15;
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -2368,8 +2368,8 @@ CLLocationCoordinate2D randomWorldCoordinate() {
     }
 }
 
-- (MGLUserLocationStyle)mapViewStyleForDefaultUserLocationAnnotationView:(MGLMapView *)mapView {
-    MGLUserLocationStyle style;
+- (MGLUserLocationAnnotationViewStyle)mapViewStyleForDefaultUserLocationAnnotationView:(MGLMapView *)mapView {
+    MGLUserLocationAnnotationViewStyle style;
     
     style.approximateHaloFillColor = UIColor.redColor;
     style.approximateHaloOpacity = 0.15;
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -2368,25 +2368,6 @@ CLLocationCoordinate2D randomWorldCoordinate() {
     }
 }
 
-- (MGLUserLocationAnnotationViewStyle *)mapViewStyleForDefaultUserLocationAnnotationView:(MGLMapView *)mapView {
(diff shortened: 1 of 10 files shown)


platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -313,6 +313,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)mapViewDidStopLocatingUser:(MGLMapView *)mapView;
 
+- (MGLUserLocationStyle)mapViewStyleForUserLocationView:(MGLMapView *)mapView;
```
1ec5, line 316: MGLUserLocationStyle appears to be specific to MGLFaux3DUserLocationAnnotationView. If the developer provides a custom view in `-mapView:viewForAnnotation:`, the return value of this method is ignored. Is there a more generic way to make MGLFaux3DUserLocationAnnotationView more stylable? How about a simple delegate method that says the view will get updated based on a new location? Or more generically, a method that says any annotation view will be updated by the map view?

fabian-guerra: You are right. This delegate method will be ignored if the users implements a custom location annotation view. We use a single annotation view to render four types of visual representations. Using a styling object was IMO is the best tradeoff to handle this. I am not sure we may want to change rendering individually each visual representation for Hydrogen.

Perhaps changing the name to `mapViewStyleForDefaultUserLocationAnnotationView:` may sound more explicit?

platform/ios/app/MBXViewController.m
```diff
@@ -2366,6 +2366,25 @@ - (void)mapView:(nonnull MGLMapView *)mapView didChangeLocationManagerAuthorizat
     }
 }
 
+- (MGLUserLocationStyle)mapViewStyleForUserLocationView:(MGLMapView *)mapView {
```
captainbarbosa, line 2369: How would you envision this working if the developer wanted to dynamically change the default user location annotation color? When we make an `ios-sdk-examples` entry for this, we should highlight this kind of use case.

platform/ios/src/MGLMapViewDelegate.h
captainbarbosa, line 316: This method should be documented. If we don't want the documentation to appear in the API guides, prepend the documentation entry with `:nodoc:`.

platform/ios/src/MGLMapView.h
```diff
@@ -133,6 +133,25 @@ FOUNDATION_EXTERN MGL_EXPORT const MGLMapViewPreferredFramesPerSecond MGLMapView
 FOUNDATION_EXTERN MGL_EXPORT MGLExceptionName const MGLMissingLocationServicesUsageDescriptionException;
 FOUNDATION_EXTERN MGL_EXPORT MGLExceptionName const MGLUserLocationAnnotationTypeException;
 
+/**
+ A structure containing information about the default User Location view style.
+ */
+typedef struct __attribute__((objc_boxable)) MGLUserLocationStyle {
```
captainbarbosa, line 139: I would encourage documenting each of these properties so that the developer has a good sense of what parts of the user location annotation are changing - some things like the `approximate` properties are new and perhaps not widely understood yet.

platform/ios/src/MGLMapViewDelegate.h
fabian-guerra: This was a work in progress. I added the documentation.

platform/ios/src/MGLMapView.h
fabian-guerra: I added the documentation.

platform/ios/src/MGLUserLocationAnnotationViewStyle.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+@property (nonatomic) UIColor *puckArrowFillColor;
+/**
+ The puck's halo fill color.
+ */
+@property (nonatomic) UIColor *haloFillColor;
+/**
+ The approximate's view halo fill color.
+ */
+@property (nonatomic) UIColor *approximateHaloFillColor;
```
captainbarbosa, line 38: Should we note that any properties for the approximate view will only work in iOS 14+ when the user has enabled approximate location?

captainbarbosa (approved):
Thanks for adding additional documentation, this is very helpful. I have two minor concerns but otherwise this looks good.

PR #411, opened 2020-09-03
Fix map stuttering by switching render call to use setNeedsDisplay

### LINKED ISSUE #350: On iOS 14 MGLAnnotationView slows down map significantly
The addition of even one custom MGLAnnotationView make the map unusable. The attached image shows the profiler where the mapbox rendering call is taking the most of the cpu. There is only a basic map showing one custom annotation. On iOS 13 this runs butter smooth. The Map is displayed via a UIViewRepresentable in SwiftUI.

image

### Steps to reproduce

 1. On iOS 14, create a MGLMapView and display one custom MGLAnnotationView.

### Expected behavior
Map runs smooth

### Actual behavior
Map is unusable slow

### Configuration
The used custom MGLAnnotationView:

```swift
class CustomAnnotationView: MGLAnnotationView {
    override func layoutSubviews() {
        super.layoutSubviews()
        layer.cornerRadius = bounds.width / 2
        layer.borderWidth = 2
        layer.borderColor = UIColor.white.cgColor
    }
}
```
**Mapbox SDK version:** 6.0.0
**iOS versions:** iPad OS 14 / iOS 13.6
**Device/simulator models:** iPad Pro 1st Gen (iOS 14 device) / iPhone Xs (iOS 13 device)
**Xcode version:** 12.0 b2

julianrex:
This PR fixes #350, where the map "stutters" while being animated. The issue is reproducible in iOS 14 beta 2 onwards (as of 2020/9/3).

The PR supersedes changes made for similar bug first reported in https://github.com/mapbox/mapbox-gl-native/issues/14232 (`FB5350728`) and:

- switches from calling `-[GLKView display]` directly (via the `CADisplayLink` call), to calling `-[GLKView setNeedsDisplay]`,
- adds development `os_signpost`s for performance analysis,
- adds a few debug options to the development app `Mapbox GL`.

### Background

This issue is seen when enabling `CAEAGLLayer.presentsWithTransaction` and calling `-[GLKView display]` from a `CADisplayLink`. `presentsWithTransaction` is used to synchronize the `UIView` annotations to the map. The stutter that is seen is the CPU blocking with `wait for next drawable`. The issue can also be reproduced with a blank map (essentially just a `glClear`).

This effect can be mitigated by setting `preferredFramesPerSecond` to 30. 

This issue has been reported as `FB8549369`.

### DIFF
--- a/platform/ios/src/MGLMapView+OpenGL.mm
+++ b/platform/ios/src/MGLMapView+OpenGL.mm
@@ -148,6 +148,7 @@ void MGLMapViewOpenGLImpl::deleteView() {
 }
 
 #ifdef MGL_RECREATE_GL_IN_AN_EMERGENCY
+// TODO: Fix or remove
 // See https://github.com/mapbox/mapbox-gl-native/issues/14232
 void MGLMapViewOpenGLImpl::emergencyRecreateGL() {
     auto& resource = getResource<MGLMapViewOpenGLRenderableResource>();
--- a/platform/ios/src/MGLMapView+OpenGL.mm
+++ b/platform/ios/src/MGLMapView+OpenGL.mm
@@ -100,34 +100,14 @@ void MGLMapViewOpenGLImpl::setPresentsWithTransaction(const bool value) {
 
 void MGLMapViewOpenGLImpl::display() {
     auto& resource = getResource<MGLMapViewOpenGLRenderableResource>();
-
-    // See https://github.com/mapbox/mapbox-gl-native/issues/14232
-    // glClear can be blocked for 1 second. This code is an "escape hatch",
-    // an attempt to detect this situation and rebuild the GL views.
-    if (mapView.enablePresentsWithTransaction && resource.atLeastiOS_12_2_0) {
-        CFTimeInterval before = CACurrentMediaTime();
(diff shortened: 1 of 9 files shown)


platform/ios/src/MGLMapView+OpenGL.mm
```diff
@@ -168,6 +148,7 @@ void bind() override {
 }
 
 #ifdef MGL_RECREATE_GL_IN_AN_EMERGENCY
+// TODO: Fix or remove
 // See https://github.com/mapbox/mapbox-gl-native/issues/14232
 void MGLMapViewOpenGLImpl::emergencyRecreateGL() {
```
nishant-karajgikar, line 153: I'm guessing we should never be getting into this situation where we need to recreate the glView anymore?

nishant-karajgikar (approved):

platform/ios/src/MGLMapView+OpenGL.mm
julianrex: 🤞 
I did test this function, but it's currently broken, and it was never a good option anyway, and an even worse one for this particular issue. Let's remove in a future PR.

Issue #217, opened 2020-03-17
Add a flag to pause network requests


julianrex:
This issue is for tracking the continuation/porting of https://github.com/mapbox/mapbox-gl-native/pull/15650. See https://github.com/mapbox/mapbox-gl-native/issues/7443 for the original issue.

@fabian-guerra @knov let's evaluate if this is still needed.

VitaliiTokar:
Hello my name is Vitalii Tokarenko) I love human

julianrex:
See https://github.com/mapbox/mapbox-gl-native-ios/issues/309 for a recent duplicate, and a possible option: https://github.com/trailbehind/mapbox-gl-native-ios/pull/1

PR #416, opened 2020-09-03
Add a flag to enforce using catched tiles

fabian-guerra:
Fixes https://github.com/mapbox/mapbox-gl-native-ios/issues/217


julianrex (approved):
Thank you!

Issue #167, opened 2020-02-18
testContentInsetCenter & testContentInsetOrnaments fail on iPhone X


julianrex:
The following unit tests are failing on an iPhone X simulator in `release-unicorn`:

- `testContentInsetCenter`
    At first glance, this appears to be an assumption in the test, that the top and bottom content insets are equal.

- `testContentInsetOrnaments`
    Although `contentInset` is set to 0, `automaticallyAdjustsContentInset` is `YES`, so will be reset to the safe insets (IIRC). However, even if this is set to `NO` the ornament constraints are still relative to the safe insets.

EDIT: Worth noting that these tests pass on an iPhone 8, which happens to be the device used in CI, hence why we haven't seen this fail PRs.

/cc @fabian-guerra @astojilj

julianrex:
cc @jmkiley

1ec5:
This issue reproduces in the wild: mapbox/mapbox-navigation-ios#2353.

PR #420, opened 2020-09-04
Fix content insets when automaticallyAdjustsScrollViewInsets is set to NO

fabian-guerra:
Fixes https://github.com/mapbox/mapbox-gl-native-ios/issues/167

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -932,6 +932,10 @@ public:
     [constraints addObjectsFromArray:updatedConstraints];
 }
 
+// Convenience method to check if a user has enabled automatically adjust content
+// insets. Currently users can use MGLMapView.automaticallyAdjustContentInset or
+// UIViewController.automaticallyAdjustsScrollViewInsets
+// TODO: Remove when UIViewController.automaticallyAdjustsScrollViewInsets is removed
 - (BOOL)hasAutomaticallyAdjustContentInset {
     BOOL automaticallyAdjustContentInset;
     if (_automaticallyAdjustContentInsetHolder) {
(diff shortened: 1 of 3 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -941,10 +932,54 @@ - (void)updateConstraintsForOrnament:(UIView *)view
     [constraints addObjectsFromArray:updatedConstraints];
 }
 
+- (BOOL)hasAutomaticallyAdjustContentInset {
```
julianrex, line 935: nit: I'm having trouble understanding the meaning of the method name.

Can we match the similar view controller method? `automaticallyAdjustsContentInsets`

depending on context?

EDIT: Oh, I see `- (BOOL)automaticallyAdjustsContentInset` - I'm more confused now 😕

platform/ios/src/MGLMapView.mm
```diff
@@ -941,10 +932,54 @@ - (void)updateConstraintsForOrnament:(UIView *)view
...
+    return automaticallyAdjustContentInset;
+}
+
+- (NSLayoutYAxisAnchor *)safeTopAnchor {
+    if ([self hasAutomaticallyAdjustContentInset]) {
+        return self.mgl_safeTopAnchor;
+    } else {
+        return self.topAnchor;
+    }
+}
```
julianrex, line 953: Another confusing one for me - is there anyway to combine `safeTapAnchor` and `mgl_safeTopAnchor`?

julianrex (approved):
Just some confusion with naming here on my part. Any chance of consolidating here?

platform/ios/src/MGLMapView.mm
fabian-guerra: `hasAutomaticallyAdjustContentInset` is an internal convenience method to avoid code duplication. Since we have two ways to check if we disabled adjusting automatically content insets: using `UIViewController.automaticallyAdjustsScrollViewInsets` (deprecated) and `MGLMapView.automaticallyAdjustContentInset` I am using this method to check these options.

platform/ios/src/MGLMapView.mm
fabian-guerra: What this method is doing is check if it has automatically adjust content insets. If it has then it uses an anchor that considers safe areas which is `self.mgl_safeTopAnchor` that is in a category, if it is not then uses the current view anchor without considering safe areas.

PR #427, opened 2020-09-10
Approximate Location "Puck" logic updates

neelmistry94:
This alters the logic for how the approximate location puck is draw. 

These changes allow the puck to be seen at an appropriate size at all zoom levels by doing the following:

- Draws/creates the approximate ring before updating the ring
- Passes zoom level when accounting ring size so it is relative to the current map's zoom level

### DIFF
--- a/platform/ios/src/MGLFaux3DUserLocationAnnotationView.mm
+++ b/platform/ios/src/MGLFaux3DUserLocationAnnotationView.mm
@@ -1,6 +1,7 @@
 #import "MGLFaux3DUserLocationAnnotationView.h"
 
 #import "MGLMapView.h"
+#import "MGLMapView_Private.h"
 #import "MGLMapViewDelegate.h"
 #import "MGLUserLocation.h"
 #import "MGLUserLocationHeadingIndicator.h"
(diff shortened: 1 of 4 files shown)


fabian-guerra (changes requested):
Is it possible to rollback files to Objc? I don't think we need Objc++ files (`.mm`)

neelmistry94:
> Is it possible to rollback files to Objc? I don't think we need Objc++ files (`.mm`)

I worked with @julianrex on this. I believe there is a compiler error if we leave it as objc.

platform/ios/src/MGLFaux3DUserLocationAnnotationView.mm
```diff
@@ -1,6 +1,7 @@
 #import "MGLFaux3DUserLocationAnnotationView.h"
 
 #import "MGLMapView.h"
```
fabian-guerra, line 3: Please remove this import.

fabian-guerra (approved):
There's one minor comment. Otherwise LGTM

fabian-guerra:
I've realized that the issue is due to the fact that in the map's private header we import a c++ header, this requires importing classes to switch to Objc++ even tho no C++ is used.

platform/ios/src/MGLFaux3DUserLocationAnnotationView.mm
neelmistry94: we moved one function to map view private which is why we included it.

fabian-guerra: Right. That's why you don't have to import `MGLMapView.h` when you have `MGLMapView_Private.h`

neelmistry94: gotcha, I can remove it

captainbarbosa:
@neelmistry94 can you please add an entry to the changelog as a follow-up PR?

fabian-guerra:
@captainbarbosa this doesn't require a changelog entry because its fixing an issue in an unreleased feature.

PR #431, opened 2020-09-14
Allow for transitive dependencies

frederoni:
Allow for transitive dependencies.

Currently when we are developing the next version of MME, the dependency tree cannot be resolved because MME in this SDK is pinned to a specific version.
```
[!] CocoaPods could not find compatible versions for pod "MapboxMobileEvents":
  In Podfile:
    Mapbox-iOS-SDK (~> 6.1.0) was resolved to 6.1.0, which depends on
      MapboxMobileEvents (= 0.10.2)
```

but I don't see any reason why we cannot allow for transitive dependencies so I'm adding the tadpole operator here.

cc @mapbox/maps-ios


julianrex:
This was related to https://github.com/mapbox/mapbox-events-ios/issues/236, getting beta.1 out the door.

We're going to merge this, and make Xcode 11.1 / 11.3.1 builds not-required checks. However, we would like a quick fix for `236` 😄

julianrex (approved):

PR #432, opened 2020-09-14
Update to application state management and associated view setup and rendering

julianrex:
This PR is essentially a port/continuation of https://github.com/mapbox/mapbox-gl-native/pull/14222 in this repo, and addresses a number of issues in https://github.com/mapbox/mapbox-gl-native/issues/13821.

It adds:
- Update application state & rendering management on resigning/become active, entering the background/foreground, window changes. I've tried to simplify each call, as there are a number of paths that an application can take. 
- A private `MGLApplication` protocol to the SDK, allowing us to mock application events for testing. This also paves the way for removing `UIApplication.shared` references from the SDK.
- Adds integration tests to test rendering correctly starts/stop/pauses/resumes during application lifecycle events
- Removes the `emergencyRecreateGL` method, that has been defunct for a while.
- Adds an undocumented property (`MGLMapView.renderingInInactiveStateEnabled`) to disable rendering when the application resigns active. This should be considered prone to change.

### DIFF
--- a/platform/ios/Integration Tests/MGLIntegrationTestCase.m
+++ b/platform/ios/Integration Tests/MGLIntegrationTestCase.m
@@ -38,6 +38,12 @@
     return newTestSuite;
 }
 
+- (void)invokeTest {
+    @autoreleasepool {
+        [super invokeTest];
+    }
+}
+
 - (void)setUp {
     [super setUp];
 
--- a/platform/ios/Integration Tests/MGLIntegrationTestCase.m
+++ b/platform/ios/Integration Tests/MGLIntegrationTestCase.m
@@ -38,11 +38,11 @@
     return newTestSuite;
 }
 
-- (void)invokeTest {
-    @autoreleasepool {
-        [super invokeTest];
-    }
-}
+//- (void)invokeTest {
+//    @autoreleasepool {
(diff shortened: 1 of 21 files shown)


platform/ios/Integration Tests/MGLIntegrationTestCase.m
```diff
@@ -38,6 +38,12 @@ + (XCTestSuite*)defaultTestSuite {
     return newTestSuite;
 }
 
+//- (void)invokeTest {
```
fabian-guerra, line 41: nit: Please remove unused code.

platform/ios/src/MGLMapView.mm
```diff
@@ -1094,7 +1133,7 @@ - (void)renderSync
 
             if (after - before >= 1.0) {
                 // See https://github.com/mapbox/mapbox-gl-native/issues/14232
-                // and https://github.com/mapbox/mapbox-gl-native-ios/issues/350
+                // and https://github.com/mapbox/xmapbox-gl-native-ios/issues/350
```
fabian-guerra, line 1136: nit: please remove the additional `x`

fabian-guerra (approved):
LGTM. Some minor comments.

PR #438, opened 2020-09-14
Update to gl-native 5.0.0

julianrex:
Updates gl-native to version 5.0.0. 

Notable changes include:
- Fix local glyph rasterizer fallback on Apple devices by appending the global cascade fallback list in case provided font is invalid. (695)
- Fix evaluation of symbol layer properties that use 'within' and/or 'distance' expressions. (736)
- During animation, avoid tile pyramid creation with an unexpected covering zoom level and hence avoid artifacts like symbol flashing. (760)

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,2 +1,2 @@
-binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps-gl-core/mapbox-ios-sdk-gl-core-static.json" "4.0.0"
-github "mapbox/mapbox-events-ios" "v0.10.2"
+binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps-gl-core/mapbox-ios-sdk-gl-core-static.json" "5.0.0"
+github "mapbox/mapbox-events-ios" "v0.10.3"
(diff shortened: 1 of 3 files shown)


Cartfile.resolved
```diff
@@ -1,2 +1,2 @@
-binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps-gl-core/mapbox-ios-sdk-gl-core-static.json" "4.0.0"
-github "mapbox/mapbox-events-ios" "v0.10.2"
+binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps-gl-core/mapbox-ios-sdk-gl-core-static.json" "5.0.0"
+github "mapbox/mapbox-events-ios" "v0.10.3"
```
fabian-guerra, line 2: shouldn't this be `v0.10.2`?

julianrex: @frederoni changed the operator to `~>`, so this is expected. It will mean that 11.1 / 11.3.1 builds will fail - but that's ok at the moment.

fabian-guerra (approved):

PR #467, opened 2020-09-18
Address crash when localFontFamilyName is nil

jmkiley:
This PR addresses a crash that can occur when `MGLIdeographicFontFamilyName` is set to `NO`.

Thanks for the fix @alexshalamov.


julianrex (approved):

alexshalamov:
@jmkiley similar fix is required for [Snaphostter](https://github.com/mapbox/mapbox-gl-native-ios/blob/master/platform/darwin/src/MGLMapSnapshotter.mm#L728)

PR #501, opened 2020-10-09
Add compass image as a sublayer

### LINKED ISSUE #500: Call to setCamera on MGLMapView is causing too many layoutSubviews updates on iOS 14.
Starting from `iOS 14` calls to animated camera changes using `MGLMapView.setCamera(_:withDuration:animationTimingFunction:)` and similar methods cause too many `layoutSubviews` updates. Such behavior cannot be reproduced on `iOS 13` and earlier. It's also not possible to reproduce behavior on `iOS 14` when setting camera without animation when using `setCamera(_:animated:)`.

### Steps to reproduce

Consider such code:

```swift
import UIKit
import Mapbox

class ViewController: UIViewController {
    
    var mapView: MGLMapView?
    var timer: Timer?
    let tag = "[Camera test]"
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        testCamera()
    }
    
    func testCamera() {
        mapView = MGLMapView(frame: view.bounds)
        mapView?.autoresizingMask = [.flexibleWidth, .flexibleHeight]
        mapView?.userTrackingMode = .follow
        mapView?.style?.transition = MGLTransition(duration: 1.0, delay: 1.0)
        mapView?.zoomLevel = 15.0
        
        view.addSubview(mapView!)
        
        timer = Timer.scheduledTimer(withTimeInterval: 5.0, repeats: true) { timer in
            if let centerCoordinate = self.mapView?.centerCoordinate {
                let camera = MGLMapCamera(lookingAtCenter: centerCoordinate,
                                          altitude: CLLocationDistance(Int.random(in: 0...1000)),
                                          pitch: CGFloat(Int.random(in: 0...45)),
                                          heading: CLLocationDirection(Int.random(in: 0...360)))
                let animationTimingFunction = CAMediaTimingFunction(name: .linear)
                NSLog("\(self.tag) Started animation.")
                self.mapView?.setCamera(camera, withDuration: 3.0, animationTimingFunction: animationTimingFunction, completionHandler: {
                    NSLog("\(self.tag) Finished animation.")
                })
            }
        }
    }
    
    override func viewDidLayoutSubviews() {
        super.viewDidLayoutSubviews()
        
        NSLog("\(tag) \(#function)")
    }
}
```

After running this code on `iOS 14` `UIViewController.viewDidLayoutSubviews` will be called too many times within really short period of time. For example: 
```
2020-10-07 09:53:52.263605-0700 MapboxCocoaPodsTest[4665:95349] [Camera test] Started animation.
2020-10-07 09:53:52.295483-0700 MapboxCocoaPodsTest[4665:95349] [Camera test] viewDidLayoutSubviews()
... 96 more times
2020-10-07 09:53:55.283449-0700 MapboxCocoaPodsTest[4665:95349] [Camera test] viewDidLayoutSubviews()
2020-10-07 09:53:55.283719-0700 MapboxCocoaPodsTest[4665:95349] [Camera test] Finished animation.
```

### Expected behavior

Since client apps might override either `UIViewController.viewDidLayoutSubviews` or `UIView.layoutSubviews` and run additional UI tasks there it might cause UI lags. This issue causes issues described in Navigation iOS SDK [here](https://github.com/mapbox/mapbox-navigation-ios/issues/2671) and [here](https://github.com/mapbox/mapbox-navigation-ios/issues/2658).

### Actual behavior

`UIViewController.viewDidLayoutSubviews` or `UIView.layoutSubviews` is called too often when setting camera on `iOS 14`.

### Configuration

**Mapbox SDK versions:** `6.2.1`
**iOS/macOS versions:**  `iOS 14.0`
**Device/simulator models:** `iPhone 11 iOS Simulator`
**Xcode version:** `Xcode 12.0`

julianrex:
Fixes #500 

Changing the transform on the compass button triggers `-[UIView layoutSubviews]` on iOS 14. This PR add a sublayer for the compass image, keeping the original UIView's transform intact.

The original issue in #500 is not fully resolved however, as the scale bar causes a similar issue (to a lesser extent).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,7 +19,7 @@ workflows:
       - ios-build:
           matrix:
             parameters:
-              xcode: ["11.1.0", "12.0.0"]
+              xcode: ["11.7.0", "12.0.0"]
               buildtype: ["Debug", "Release"]
       - ios-release-template:
           xcode: "12.0.0"
@@ -57,7 +57,7 @@ workflows:
       - ios-build:
           matrix:
             parameters:
-              xcode: ["11.3.1", "11.7.0", "12.2.0"]
+              xcode: ["11.1.0", "11.3.1", "12.2.0"]
               buildtype: ["Debug", "Release"]
       - ios-sanitize-nightly:
           requires:
--- a/circle.yml
+++ b/circle.yml
@@ -19,7 +19,7 @@ workflows:
       - ios-build:
           matrix:
             parameters:
-              xcode: ["11.1.0", "11.7.0"]
+              xcode: ["11.1.0", "12.0.0"]
               buildtype: ["Debug", "Release"]
       - ios-release-template:
           xcode: "12.0.0"
@@ -57,7 +57,7 @@ workflows:
       - ios-build:
           matrix:
             parameters:
-              xcode: ["11.3.1", "11.5.0", "11.6.0", "12.0.0"]
+              xcode: ["11.3.1", "11.7.0", "12.2.0"]
               buildtype: ["Debug", "Release"]
       - ios-sanitize-nightly:
           requires:
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -4,14 +4,19 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ## 6.3.0
 
+### 🐞 Bug fixes
+
+* Partly fixed an issue on iOS 14 where `-[UIView layoutSubviews]` was being repeatedly called. ([#501](https://github.com/mapbox/mapbox-gl-native-ios/pull/501))
+* Fixed a bug with UIViews being incorrectly updated with a one frame delay. ([#483](https://github.com/mapbox/mapbox-gl-native-ios/pull/483))
+* Fixed an issue where CocoaPods users could not install the SDK when using Xcode 12. ([#482](https://github.com/mapbox/mapbox-gl-native-ios/pull/482))
+
 ### 🔧 Dependencies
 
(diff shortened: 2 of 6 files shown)


julianrex:
cc @MaximAlien

lloydsheng:
Since we have added an `imageLayer`, does `MGLCompassButton` still need to inherit `UIImageView` ?
I think there may be a better solution to this issue. In order to avoid `-[UIView layoutSubviews]` being called, Do we have to add an extra layer to the subview?

julianrex:
> Since we have added an imageLayer, does MGLCompassButton still need to inherit UIImageView ?

Good catch; because it's called "Button" I had assumed it was a `UIButton`. I'll change to a plain `UIView`.

> I think there may be a better solution to this issue. In order to avoid -[UIView layoutSubviews] being called, Do we have to add an extra layer to the subview?

Something other than the compass view needs to have its transform changed. This works for both a subview and a sublayer. I chose a sublayer since it won't play a role in auto-layout.

MaximAlien (commented):
Would it be possible to back-port current fix to version `6.2` as well?

lloydsheng (approved):

circle.yml
```diff
@@ -57,7 +57,7 @@ workflows:
       - ios-build:
           matrix:
             parameters:
-              xcode: ["11.3.1", "11.5.0", "11.6.0", "12.0.0"]
+              xcode: ["11.1.0", "11.3.1", "12.2.0"]
```
captainbarbosa, line 60: For next time, I'd maybe consider splitting this into a separate PR.

captainbarbosa (approved):
Look good, just one question - does this change any accessibility-related functionality?

julianrex:
> does this change any accessibility-related functionality?

No, this should be a visual change only.

Issue #502, opened 2020-10-09
Scalebar ornament triggers layoutSubviews.


julianrex:
Follows on from #500:

Turns out both the compass and scalebar ornaments are triggering layoutSubviews. See #501 for a fix for the compass.

For the scalebar, this is due to auto layout and the use of intrinsicContentSize.

_Originally posted by @julianrex in https://github.com/mapbox/mapbox-gl-native-ios/issues/500#issuecomment-705944850_

PR #507, opened 2020-10-13
Fix scalebar ornament triggers unnecessary layoutSubviews

lloydsheng:
Fix issue #502

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -866,6 +866,7 @@ public:
 - (void)setScaleBarPosition:(MGLOrnamentPosition)scaleBarPosition {
     MGLLogDebug(@"Setting scaleBarPosition: %lu", scaleBarPosition);
     _scaleBarPosition = scaleBarPosition;
+    _scaleBar.isOnScreenRight = (scaleBarPosition == MGLOrnamentPositionTopRight || scaleBarPosition == MGLOrnamentPositionBottomRight);
     [self installScaleBarConstraints];
 }
 
@@ -914,7 +915,7 @@ public:
 - (void)updateConstraintsForOrnament:(UIView *)view
                          constraints:(NSMutableArray *)constraints
                             position:(MGLOrnamentPosition)position
-                                size:(CGSize)size
+                     sizeConstraints:(NSArray *)sizeConstraints
                              margins:(CGPoint)margins {
     NSMutableArray *updatedConstraints = [NSMutableArray array];
     UIEdgeInsets inset = UIEdgeInsetsZero;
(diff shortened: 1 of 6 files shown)


ZiZasaurus (commented):
added log statement to verify that unnecessary layoutSubviews weren't being triggered. Looks good!

platform/ios/src/MGLMapView.mm
```diff
@@ -1061,29 +1056,56 @@ - (void)installAttributionButtonConstraints {
...
+                                                       attribute:NSLayoutAttributeWidth
+                                                      multiplier:.5
+                                                        constant:0];
+        heightConstraint = [NSLayoutConstraint constraintWithItem:_scaleBar
+                                                        attribute:NSLayoutAttributeHeight
+                                                        relatedBy:NSLayoutRelationEqual
+                                                           toItem:nil
+                                                        attribute:NSLayoutAttributeNotAnAttribute
+                                                       multiplier:1
+                                                         constant:16];
```
julianrex, line 1083: Do we have an existing constant for this?

platform/ios/src/MGLMapView.mm
```diff
@@ -1061,29 +1056,56 @@ - (void)installAttributionButtonConstraints {
...
+    NSLayoutConstraint *widthConstraint = nil;
+    NSLayoutConstraint *heightConstraint = nil;
+
+    if (view == _scaleBar) {
+        widthConstraint = [NSLayoutConstraint constraintWithItem:_scaleBar
+                                                       attribute:NSLayoutAttributeWidth
+                                                       relatedBy:NSLayoutRelationEqual
+                                                          toItem:self
+                                                       attribute:NSLayoutAttributeWidth
+                                                      multiplier:.5
```
julianrex, line 1075: Is half the width of the view essentially the maximum size possible? How does this impact running on iPad?

platform/ios/src/MGLScaleBar.mm
```diff
@@ -498,21 +464,21 @@ - (void)layoutSubviews {
...
-    CGFloat intrinsicContentHeight = self.intrinsicContentSize.height;
     CGFloat barWidth               = totalBarWidth/self.bars.count;
 
     BOOL RTL               = [self usesRightToLeftLayout];
     CGFloat halfLabelWidth = ceil(self.lastLabelWidth/2);
     CGFloat barOffset      = RTL ? halfLabelWidth : 0.0;
     
-    self.containerView.frame = CGRectMake(barOffset,
-                                          intrinsicContentHeight - MGLBarHeight,
+    CGFloat containerViewX = self.isOnScreenRight ? self.bounds.size.width - totalBarWidth - barOffset : barOffset;
```
julianrex, line 480: Does this also need the `halfLabelWidth` if on the right hand side?

julianrex (approved):
A few questions that can be addressed if appropriate, but otherwise looks good. Thank you!

ZiZasaurus (approved):

platform/ios/src/MGLMapView.mm
lloydsheng: Yes, I get the height value from https://github.com/mapbox/mapbox-gl-native-ios/pull/507/files/ab5ff48099caff30598e35b2d7a2edc49539babd#diff-6f8e414d71df96dda5dfecbd275ee6c5a35b52e337f3f941c6c132abda50ffb0L347

platform/ios/src/MGLMapView.mm
lloydsheng: It may be a bit long on the iPad, but I think it's acceptable 😂

platform/ios/src/MGLScaleBar.mm
lloydsheng: It needs the `barOffset`(calculated by `halfLabelWidth `) for layout direction right to left on either side. Otherwise, labels will display like this 👇
image

PR #517, opened 2020-10-22
[ios] Add glyphs rasterization options

lloydsheng:
Add `MGLGlyphsRasterizationMode` indicates how the map view load glyphs.

### DIFF
--- a/platform/darwin/src/MGLMapSnapshotter.mm
+++ b/platform/darwin/src/MGLMapSnapshotter.mm
@@ -14,7 +14,7 @@
 #import "MGLStyle_Private.h"
 #import "MGLAttributionInfo_Private.h"
 #import "MGLLoggingConfiguration_Private.h"
-#import "MGLRendererConfiguration.h"
+#import "MGLRendererConfiguration_Private.h"
 #import "MGLMapSnapshotter_Private.h"
 
 #if TARGET_OS_IPHONE || TARGET_OS_SIMULATOR
@@ -725,24 +725,9 @@ NSArray<MGLAttributionInfo *> *MGLAttributionInfosFromAttributions(mbgl::MapSnap
                    .withAssetPath(NSBundle.mainBundle.resourceURL.path.UTF8String);
 
     // Create the snapshotter
-    mbgl::optional<std::string> localFontFamilyName = config.localFontFamilyName ? mbgl::optional<std::string>(std::string(config.localFontFamilyName.UTF8String)) : mbgl::nullopt;
-    mbgl::GlyphsRasterizationOptions glyphsRasterizationOptions;
-    glyphsRasterizationOptions.fontFamily = localFontFamilyName;
-    switch (config.glyphsRasterizationMode) {
-        case MGLNoGlyphsRasterizedLocally:
-            glyphsRasterizationOptions.rasterizationMode = mbgl::GlyphsRasterizationMode::NoGlyphsRasterizedLocally;
-            break;
-        case MGLIdeographsRasterizedLocally:
-            glyphsRasterizationOptions.rasterizationMode = mbgl::GlyphsRasterizationMode::IdeographsRasterizedLocally;
-            break;
-        case MGLAllGlyphsRasterizedLocally:
-            glyphsRasterizationOptions.rasterizationMode = mbgl::GlyphsRasterizationMode::AllGlyphsRasterizedLocally;
-            break;
-    }
-
     _delegateHost = std::make_unique<MGLMapSnapshotterDelegateHost>(self);
     _mbglMapSnapshotter = std::make_unique<mbgl::MapSnapshotter>(
-                                                                 size, pixelRatio, resourceOptions, *_delegateHost, glyphsRasterizationOptions);
+                                                                 size, pixelRatio, resourceOptions, *_delegateHost, config.glyphsRasterizationOptions);
     
     _mbglMapSnapshotter->setStyleURL(std::string(options.styleURL.absoluteString.UTF8String));
     
--- a/platform/darwin/src/MGLMapSnapshotter.mm
+++ b/platform/darwin/src/MGLMapSnapshotter.mm
@@ -726,9 +726,23 @@ NSArray<MGLAttributionInfo *> *MGLAttributionInfosFromAttributions(mbgl::MapSnap
 
     // Create the snapshotter
     mbgl::optional<std::string> localFontFamilyName = config.localFontFamilyName ? mbgl::optional<std::string>(std::string(config.localFontFamilyName.UTF8String)) : mbgl::nullopt;
+    mbgl::GlyphsRasterizationOptions glyphsRasterizationOptions;
+    glyphsRasterizationOptions.fontFamily = localFontFamilyName;
+    switch (config.glyphsRasterizationMode) {
+        case MGLNoGlyphsRasterizedLocally:
+            glyphsRasterizationOptions.rasterizationMode = mbgl::GlyphsRasterizationMode::NoGlyphsRasterizedLocally;
+            break;
+        case MGLIdeographsRasterizedLocally:
+            glyphsRasterizationOptions.rasterizationMode = mbgl::GlyphsRasterizationMode::IdeographsRasterizedLocally;
+            break;
+        case MGLAllGlyphsRasterizedLocally:
+            glyphsRasterizationOptions.rasterizationMode = mbgl::GlyphsRasterizationMode::AllGlyphsRasterizedLocally;
+            break;
+    }
+
     _delegateHost = std::make_unique<MGLMapSnapshotterDelegateHost>(self);
     _mbglMapSnapshotter = std::make_unique<mbgl::MapSnapshotter>(
-                                                                 size, pixelRatio, resourceOptions, *_delegateHost, localFontFamilyName);
+                                                                 size, pixelRatio, resourceOptions, *_delegateHost, glyphsRasterizationOptions);
     
     _mbglMapSnapshotter->setStyleURL(std::string(options.styleURL.absoluteString.UTF8String));
     
--- a/platform/darwin/src/MGLRendererConfiguration.h
+++ b/platform/darwin/src/MGLRendererConfiguration.h
@@ -5,12 +5,14 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// Indicates how the map view load glyphs.
 typedef NS_ENUM(NSUInteger, MGLGlyphsRasterizationMode) {
+    /// The MGLGlyphsRasterizationMode was unset .
+    MGLGlyphsRasterizationModeNone,
     /// No glyphs are rasterized locally. All glyphs are loaded from the server.
-    MGLNoGlyphsRasterizedLocally = 0,
+    MGLGlyphsRasterizationModeIdeographsRasterizedLocally,
     /// Ideographs are rasterized locally, and they are not loaded from the server.
-    MGLIdeographsRasterizedLocally,
+    MGLGlyphsRasterizationModeNoGlyphsRasterizedLocally,
     /// All glyphs are rasterized locally. No glyphs are loaded from the server.
-    MGLAllGlyphsRasterizedLocally
+    MGLGlyphsRasterizationModeAllGlyphsRasterizedLocally
 };
 
 /**
--- a/platform/darwin/src/MGLRendererConfiguration.h
+++ b/platform/darwin/src/MGLRendererConfiguration.h
@@ -5,11 +5,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// Indicates how the map view load glyphs.
 typedef NS_ENUM(NSUInteger, MGLGlyphsRasterizationMode) {
-    /// The MGLGlyphsRasterizationMode was unset .
+    /// The MGLGlyphsRasterizationMode was unset.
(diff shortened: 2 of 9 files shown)


julianrex (approved):
Looks good! thank you

platform/darwin/src/MGLMapSnapshotter.mm
```diff
@@ -726,9 +726,23 @@ - (void)configureWithOptions:(MGLMapSnapshotOptions *)options {
 
     // Create the snapshotter
     mbgl::optional<std::string> localFontFamilyName = config.localFontFamilyName ? mbgl::optional<std::string>(std::string(config.localFontFamilyName.UTF8String)) : mbgl::nullopt;
+    mbgl::GlyphsRasterizationOptions glyphsRasterizationOptions;
```
alexshalamov, line 729: If `MGLIdeographicFontFamilyName` is provided and it is set to `NO`, and `MGLGlyphsRasterizationMode` is set to something other than `MGLNoGlyphsRasterizedLocally`, maybe we should print error and fallback to `NoGlyphsRasterizedLocally` option.

Maybe worth adding test for that combination?

platform/darwin/src/MGLMapSnapshotter.mm
```diff
@@ -726,9 +726,23 @@ - (void)configureWithOptions:(MGLMapSnapshotOptions *)options {
...
+    glyphsRasterizationOptions.fontFamily = localFontFamilyName;
```
alexshalamov, line 730: if localFontFamilyName is `nullopt`, we should fallback to `MGLNoGlyphsRasterizedLocally`

platform/darwin/src/MGLRendererConfiguration.m
```diff
@@ -63,4 +64,20 @@ - (BOOL)perSourceCollisionsWithInfoDictionaryObject:(nullable id)infoDictionaryO
...
 }
 
+- (MGLGlyphsRasterizationMode)glyphsRasterizationMode {
+    id infoDictionaryObject = [NSBundle.mainBundle objectForInfoDictionaryKey:MGLGlyphsRasterizationModeKey];
+    return [self glyphsRasterizationModeWithInfoDictionaryObject:infoDictionaryObject];
+}
+
+- (MGLGlyphsRasterizationMode)glyphsRasterizationModeWithInfoDictionaryObject:(id)infoDictionaryObject {
+    if (!infoDictionaryObject || ![infoDictionaryObject isKindOfClass:[NSString class]]) {
+        return MGLNoGlyphsRasterizedLocally;
```
alexshalamov, line 74: I think it should return something that represents 'unset' value, `nil` cc/ @julianrex?

With this PR, if user don't set `MGLIdeographicFontFamilyName` and `MGLGlyphsRasterizationMode`, `glyphsRasterizationModeWithInfoDictionaryObject` would return `MGLNoGlyphsRasterizedLocally`, thus, we will not render CJK characters locally. That will be a breaking change.

platform/ios/src/MGLMapView.mm
```diff
@@ -557,7 +557,21 @@ - (void)commonInit
...
+        case MGLNoGlyphsRasterizedLocally:
+            glyphsRasterizationOptions.rasterizationMode = mbgl::GlyphsRasterizationMode::NoGlyphsRasterizedLocally;
+            break;
+        case MGLIdeographsRasterizedLocally:
+            glyphsRasterizationOptions.rasterizationMode = mbgl::GlyphsRasterizationMode::IdeographsRasterizedLocally;
+            break;
+        case MGLAllGlyphsRasterizedLocally:
+            glyphsRasterizationOptions.rasterizationMode = mbgl::GlyphsRasterizationMode::AllGlyphsRasterizedLocally;
+            break;
+    }
```
alexshalamov, line 572: I think this block could be shared between snapshotter and mapview

platform/darwin/src/MGLMapSnapshotter.mm
julianrex: Definitely.

platform/darwin/src/MGLMapSnapshotter.mm
julianrex: +1

platform/darwin/src/MGLRendererConfiguration.m
julianrex: Excellent point. Rather than `nil` (which would make this return type a bit weird), what if we add `MGLGlyphsRasterizationModeNone`?

platform/darwin/src/MGLRendererConfiguration.h
```diff
@@ -3,6 +3,16 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
+/// Indicates how the map view load glyphs.
+typedef NS_ENUM(NSUInteger, MGLGlyphsRasterizationMode) {
+    /// No glyphs are rasterized locally. All glyphs are loaded from the server.
+    MGLNoGlyphsRasterizedLocally = 0,
+    /// Ideographs are rasterized locally, and they are not loaded from the server.
+    MGLIdeographsRasterizedLocally,
+    /// All glyphs are rasterized locally. No glyphs are loaded from the server.
+    MGLAllGlyphsRasterizedLocally
```
julianrex, line 13: These enums should have the same prefix, e.g. for Swift consumption. Although it's a mouthful, how about:

```
MGLGlyphsRasterizationModeNone
MGLGlyphsRasterizationModeIdeographsRasterizedLocally
MGLGlyphsRasterizationModeNoGlyphsRasterizedLocally
MGLGlyphsRasterizationModeAllGlyphsRasterizedLocally
```

julianrex (changes requested):

lloydsheng:
@alexander-kulikovskii @julianrex Thanks for your review 🙇 .

platform/darwin/src/MGLRendererConfiguration.mm
```diff
@@ -0,0 +1,119 @@
...
+
+- (MGLGlyphsRasterizationMode)glyphsRasterizationModeWithInfoDictionaryObject:(id)infoDictionaryObject {
+    if (!infoDictionaryObject || ![infoDictionaryObject isKindOfClass:[NSString class]]) {
+        return MGLGlyphsRasterizationModeNone;
+    }
+    NSDictionary *nameOptionMap = @{@"MGLNoGlyphsRasterizedLocally":@(MGLGlyphsRasterizationModeIdeographsRasterizedLocally),
+                                    @"MGLIdeographsRasterizedLocally":@(MGLGlyphsRasterizationModeNoGlyphsRasterizedLocally),
+                                    @"MGLAllGlyphsRasterizedLocally":@(MGLGlyphsRasterizationModeAllGlyphsRasterizedLocally)};
+
+    return (MGLGlyphsRasterizationMode)[nameOptionMap[infoDictionaryObject] integerValue];
```
julianrex, line 81: For other readers: this should result in `MGLGlyphsRasterizationModeNone` if `infoDictionaryObject` is not expected.

See: https://github.com/mapbox/mapbox-gl-native-ios/pull/517/files#diff-944ba4f88de8974791fb7cf7e890a87157c51513eb6e31179983619a8312c6a6R176

julianrex (approved):

platform/darwin/src/MGLRendererConfiguration.mm
```diff
@@ -0,0 +1,119 @@
...
+
+- (mbgl::GlyphsRasterizationOptions)glyphsRasterizationOptions {
+    return [self glyphsRasterizationOptionsWithLocalFontFamilyName:self.localFontFamilyName rasterizationMode:self.glyphsRasterizationMode];
+}
+
+- (mbgl::GlyphsRasterizationOptions)glyphsRasterizationOptionsWithLocalFontFamilyName:(nullable NSString *)fontFamilyName
+                                                                    rasterizationMode:(MGLGlyphsRasterizationMode)rasterizationMode {
+    mbgl::GlyphsRasterizationOptions options;
+    if (fontFamilyName == nil) {
+        if (rasterizationMode != MGLGlyphsRasterizationModeNoGlyphsRasterizedLocally) {
```
zmiao, line 92: Is it possible that `rasterizationMode == MGLGlyphsRasterizationModeNone`?

platform/darwin/src/MGLRendererConfiguration.mm
```diff
@@ -0,0 +1,119 @@
...
+            options.rasterizationMode = mbgl::GlyphsRasterizationMode::IdeographsRasterizedLocally;
+            break;
+        case MGLGlyphsRasterizationModeNoGlyphsRasterizedLocally:
+            options.rasterizationMode = mbgl::GlyphsRasterizationMode::NoGlyphsRasterizedLocally;
+            break;
+        case MGLGlyphsRasterizationModeAllGlyphsRasterizedLocally:
+            options.rasterizationMode = mbgl::GlyphsRasterizationMode::AllGlyphsRasterizedLocally;
+            break;
+        default:
+            options.rasterizationMode = mbgl::GlyphsRasterizationMode::NoGlyphsRasterizedLocally;
```
zmiao, line 111: How about mode `MGLGlyphsRasterizationModeNone`? 
And should we print an error log in case that the mode fallback to default?

platform/darwin/src/MGLRendererConfiguration.mm
```diff
@@ -0,0 +1,119 @@
...
+    mbgl::GlyphsRasterizationOptions options;
+    if (fontFamilyName == nil) {
+        if (rasterizationMode != MGLGlyphsRasterizationModeNoGlyphsRasterizedLocally) {
+            MGLLogError(@"The `MGLIdeographicFontFamilyName` is set to `NO`, this will make `MGLGlyphsRasterizationMode` always be `MGLNoGlyphsRasterizedLocally`.");
+        }
+        options.rasterizationMode = mbgl::GlyphsRasterizationMode::NoGlyphsRasterizedLocally;
+        return options;
+    }
+    
+    options.fontFamily =  mbgl::optional<std::string>(std::string(fontFamilyName.UTF8String));
```
alexshalamov, line 99: nit: `options.fontFamily =  std::string(fontFamilyName.UTF8String);` should be enough

platform/darwin/src/MGLRendererConfiguration.mm
alexshalamov: Yes, we should always fallback to default. If `fontFamilyName` is not nil and user sets [`MGLNoGlyphsRasterizedLocally`](https://github.com/mapbox/mapbox-gl-native-ios/pull/517/files#diff-5088f83de2b6336d6e2b089e057d5dc2a6c4025d5809fe29e50d0cd307dd4992R104) or `MGLGlyphsRasterizationModeNone`, I think we should warn and fallback to `MGLIdeographsRasterizedLocally`

platform/darwin/src/MGLRendererConfiguration.mm
```diff
@@ -0,0 +1,119 @@
...
+    id infoDictionaryObject = [NSBundle.mainBundle objectForInfoDictionaryKey:MGLGlyphsRasterizationModeKey];
+    return [self glyphsRasterizationModeWithInfoDictionaryObject:infoDictionaryObject];
+}
+
+- (MGLGlyphsRasterizationMode)glyphsRasterizationModeWithInfoDictionaryObject:(id)infoDictionaryObject {
+    if (!infoDictionaryObject || ![infoDictionaryObject isKindOfClass:[NSString class]]) {
+        return MGLGlyphsRasterizationModeNone;
+    }
+    NSDictionary *nameOptionMap = @{@"MGLNoGlyphsRasterizedLocally":@(MGLGlyphsRasterizationModeIdeographsRasterizedLocally),
+                                    @"MGLIdeographsRasterizedLocally":@(MGLGlyphsRasterizationModeNoGlyphsRasterizedLocally),
```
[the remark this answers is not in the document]
julianrex: Nice catch.

platform/darwin/src/MGLRendererConfiguration.mm
```diff
@@ -0,0 +1,119 @@
...
+- (mbgl::GlyphsRasterizationOptions)glyphsRasterizationOptions {
+    return [self glyphsRasterizationOptionsWithLocalFontFamilyName:self.localFontFamilyName rasterizationMode:self.glyphsRasterizationMode];
+}
+
+- (mbgl::GlyphsRasterizationOptions)glyphsRasterizationOptionsWithLocalFontFamilyName:(nullable NSString *)fontFamilyName
+                                                                    rasterizationMode:(MGLGlyphsRasterizationMode)rasterizationMode {
+    mbgl::GlyphsRasterizationOptions options;
+    if (fontFamilyName == nil) {
+        if (rasterizationMode != MGLGlyphsRasterizationModeNoGlyphsRasterizedLocally) {
+            MGLLogError(@"The `MGLIdeographicFontFamilyName` is set to `NO`, this will make `MGLGlyphsRasterizationMode` always be `MGLNoGlyphsRasterizedLocally`.");
```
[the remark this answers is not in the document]
jmkiley: We'll ticket this out and address after beta has landed.

platform/darwin/src/MGLRendererConfiguration.mm
jmkiley: We may want to revisit and test the behavior here once beta is released.

PR #536, opened 2020-11-05
Fix #454 colors in feature attributes are misinterpreted when used in key path expressions

### LINKED ISSUE #454: Colors in feature attributes are misinterpreted when used in key path expressions
NSColor and UIColor objects used as attribute values of MGLFeature are no longer automatically premultiplied the correct number of times, so an `MGLFillStyleLayer.fillColor` expression based on a feature attribute would render features with the wrong fill color.

## Expected behavior

For convenience’s sake, [MGLFeature automatically converts attribute values](https://docs.mapbox.com/ios/api/maps/6.0.0/Protocols/MGLFeature.html#/c:objc%28pl%29MGLFeature%28py%29attributes) of type NSColor (on macOS) or UIColor (iOS) to a CSS-formatted `rgba()` string, which happens to be compatible with [color-typed paint properties](https://docs.mapbox.com/mapbox-gl-js/style-spec/types/#color) in the style specification:

> In addition to the Foundation types, you may also set an attribute to an NSColor (macOS) or UIColor (iOS), which will be converted into its CSS string representation when the feature is added to an MGLShapeSource. This can be convenient when using the attribute to supply a value for a color-typed layout or paint attribute via a key path expression.

The developer would expect this conversion to be accurate, so that the result of setting the `MGLFillStyleLayer.fillColor` property would match the normal appearance of the NSColor or UIColor that was stored in the feature attribute. For example, the following code should result in a translucent, light blue fill:

```swift
let highlightFeature = MGLPolygonFeature(coordinates: coordinates, count: UInt(points.count))
highlightFeature.attributes["fill"] = UIColor(red: 145.0 / 255.0, green: 174.0 / 255.0, blue: 255.0 / 255.0, alpha: 0.5)
let highlightSource = MGLShapeSource(identifier: "highlight", shape: highlightFeature, options: nil)
style.addSource(highlightSource)

let highlightLayer = MGLFillStyleLayer(identifier: "highlight", source: highlightSource)
highlightLayer.fillColor = NSExpression(forKeyPath: "fill")
style.addLayer(highlightLayer)
```

## Actual behavior

The code actually results in a drab gray fill.

## Diagnosis

The convenience conversion works by stuffing the NSColor or UIColor in a constant value expression and converting the expression to a JSON-formatted object:

https://github.com/mapbox/mapbox-gl-native-ios/blob/6a5db0b20282fe817a38381f47f2814a218f2382/platform/darwin/src/MGLFeature.mm#L539 https://github.com/mapbox/mapbox-gl-native-ios/blob/6a5db0b20282fe817a38381f47f2814a218f2382/platform/darwin/src/NSDictionary%2BMGLAdditions.mm#L17-L18

The conversion to JSON takes the four color channels as is without premultiplying the alpha channel into the red, green, and blue channels:

https://github.com/mapbox/mapbox-gl-native-ios/blob/6a5db0b20282fe817a38381f47f2814a218f2382/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L320-L322 https://github.com/mapbox/mapbox-gl-native-ios/blob/6a5db0b20282fe817a38381f47f2814a218f2382/platform/ios/src/UIColor%2BMGLAdditions.mm#L15-L21

From mapbox/mapbox-gl-native#8025 until #266, this conversion from NSColor/UIColor to JSON premultiplied the color, based on a historical expectation that `mbgl::Color` was always premultiplied:

https://github.com/mapbox/mapbox-gl-native-ios/blob/23b5b46cd49dd5f1267fd84f53938e6480c8001c/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L310-L312 https://github.com/mapbox/mapbox-gl-native-ios/blob/6a5db0b20282fe817a38381f47f2814a218f2382/platform/ios/src/UIColor%2BMGLAdditions.mm#L5-L13

Even if the change in #266 was correct in the usual case when converting expressions from NSExpression to JSON format, for code that uses the feature attribute convenience conversion, the change affected it twice over.

## Workaround

A workaround is to circumvent the any color-related conversions in the platform-specific SDK code by specifying a CSS-style `rgba(145.0, 174.0, 255.0, 0.5)` string directly as the attribute value:

```swift
let highlightColor = UIColor(red: 145.0 / 255.0, green: 174.0 / 255.0, blue: 255.0 / 255.0, alpha: 0.5)
let red = UnsafeMutablePointer<CGFloat>.allocate(capacity: 1)
let green = UnsafeMutablePointer<CGFloat>.allocate(capacity: 1)
let blue = UnsafeMutablePointer<CGFloat>.allocate(capacity: 1)
let alpha = UnsafeMutablePointer<CGFloat>.allocate(capacity: 1)
highlightColor.getRed(red, green: green, blue: blue, alpha: alpha)
highlightFeature.attributes["fill"] = "rgba(\(red.pointee * 255.0), \(green.pointee * 255.0), \(blue.pointee * 255.0), \(alpha.pointee))"
```

Performance-wise, this should be equivalent to the convenience conversion provided by the SDK, or even more performant in the case that the feature is reused multiple times in the same shape source or by multiple shape sources.

/cc @mapbox/maps-ios

lloydsheng:
`<changelog>Fix issue which colors in feature attributes are misinterpreted when used in key path expressions</changelog>`


platform/darwin/test/MGLExpressionTests.mm
```diff
@@ -64,7 +64,8 @@ - (void)testStringValuation {
 }
 
 - (void)testColorValuation {
-    MGLAssertConstantEqualsValue([MGLColor redColor], "rgba(255,0,0,1)"s, @"MGLColor should convert to std::string containing CSS color string.");
+    MGLAssertConstantEqualsValue([MGLColor colorWithRed:147/255.0 green:123/255.0 blue:100/255.0 alpha:1], "rgb(147, 123, 100)"s, @"MGLColor should convert to std::string containing CSS color string.");
+    MGLAssertConstantEqualsValue([MGLColor colorWithRed:147/255.0 green:123/255.0 blue:100/255.0 alpha:0.3], "rgba(147, 123, 100, 0.3)"s, @"MGLColor should convert to std::string containing CSS color string.");
```
1ec5, line 68: The `s` after the rgba string literal looks like a (preexisting) typo. Any idea if it’s necessary?

platform/ios/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * Fixed a bug with UIViews being incorrectly updated with a one frame delay. ([#483](https://github.com/mapbox/mapbox-gl-native-ios/pull/483))
 * Fixed an issue where CocoaPods users could not install the SDK when using Xcode 12. ([#482](https://github.com/mapbox/mapbox-gl-native-ios/pull/482))
 * Fixed an issue on iOS 14 where scalebar triggers unnecessary `-[UIView layoutSubviews]`. ([#507](https://github.com/mapbox/mapbox-gl-native-ios/pull/507))
+* Fixed an issue where colors in feature attributes are misinterpreted when used in key path. ([#536](https://github.com/mapbox/mapbox-gl-native-ios/pull/536))
```
1ec5, line 13:
```suggestion
* Fixed an issue where a color in a feature attribute was misinterpreted when used in a key path expression. ([#536](https://github.com/mapbox/mapbox-gl-native-ios/pull/536))
```

platform/ios/src/UIColor+MGLAdditions.mm
```diff
@@ -2,6 +2,12 @@
 
 @implementation UIColor (MGLAdditions)
 
+static inline unsigned ToByte(CGFloat f)
+{
+    f = MAX(0, MIN(f, 1)); // Clamp
```
1ec5, line 7: I think clamping the value between 0 and 1 would lead to inaccuracies on a wide color–enabled display, but then again I don’t know that mbgl configures OpenGL correctly for wide color. Probably it doesn’t, because the [`rgba`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#rgba) expression is documented as accepting only values between 0 and 1.

platform/ios/src/UIColor+MGLAdditions.mm
```diff
@@ -2,6 +2,12 @@
...
+    return (unsigned)round(f * 255);
```
1ec5, line 8: The style specification declares each component’s type as a JSON number, which can be a float. In practice, I don’t think mbgl supports deep color, so the values would get rounded anyways.

1ec5 (approved):
Two nitpicks and a couple comments for future reference.

1ec5 (commented):
Remember to test both that #125 hasn’t regressed and that the feature attributes in #454 are converted correctly.

platform/darwin/test/MGLExpressionTests.mm
lloydsheng: The `s` is necessary. I will shows error`No matching member function for call to 'get'` if it is removed.

Issue #570, opened 2020-12-08
Unusable .followWithHeading tracking mode


tomascern:
### Steps to reproduce

 1. Use custom user location annotation
 2. Set .followWithHeading user tracking mode
 3. In this case, official Mapbox Examples app is used ("Customize the user location annotation" example)
 4. Change heading while moving

### Expected behavior
Actual user location is followed with custom annotation precisely and map orientation is changed according to actual user's heading.

### Actual behavior
When the heading is being changed, user location annotation doesn't follow actual user's location. 

### Configuration

**Mapbox SDK versions:** 6.3.0
**iOS/macOS versions: 14.2
**Device/simulator models: Device - iPhone SE(1st generation), iPhone 8
**Xcode version:** XCode 12.2

![ezgif com-video-to-gif-3](https://user-images.githubusercontent.com/13505283/101535594-2ba5d580-3999-11eb-924b-6bc591ed6eaf.gif)

ZiZasaurus:
@tomascern, thank you for writing in. I was able to reproduce the behavior reported above and will be investigating further to see what may be causing it.

PR #580, opened 2021-01-05
Fix an issue where updating .followWithHeading prevents annotation's location updates

ZiZasaurus:
Submitting potential fix for `followWithHeading` issue to address https://github.com/mapbox/mapbox-gl-native-ios/issues/570

**Before change:** 
BeforeChange

**After change:** 
AfterChange


macdrevx:
@ZiZasaurus looks like you picked up a commit from some other branch as well. Was that intentional?

ZiZasaurus:
Ah, I see that! No, not intentional. How should I fix it?

Issue #562, opened 2020-11-30
[iOS] Enabling Voice Over crashes map in app when English language is not installed on device


cbrevik:
When using Mapbox along with Voice Over, the app crashes. This seems to be the case if English is not installed on the device. I have Norwegian language on my device/Voice Over. If I install English in the device menu, the crash goes away. In other words, I don't even have to enable English as the main language, it just needs to be installed.

**Edit**: As @brorhb pointed out, it doesn't have to be English, but just one of these languages:
https://github.com/mapbox/mapbox-gl-native-ios/blob/975202c07aee382f737dca28510112bee44bab57/platform/darwin/src/MGLVectorTileSource.mm#L92

The error message is:
`Exception	NSException *	"*** -[__NSPlaceholderArray initWithObjects:count:]: attempt to insert nil object from objects[0]"	0x00000002834e5860`

The same error has been noted in https://github.com/react-native-mapbox-gl/maps/issues/1113 (I use Mapbox with React Native), and I first thought it was an error with that library which wraps this one. But after further research, and trying to use the official samples from Mapbox (https://github.com/mapbox/ios-sdk-examples/), it seems to be related to the SDK here.

Video of the crash: https://youtu.be/hHjgPSoQ9Lc
Does not crash if English is installed in device settings: https://youtu.be/rOiy12W7X7s

Stack trace:
```log
Thread#0	0x00000001ac1ec878 in __exceptionPreprocess ()
#1	0x00000001c0742c50 in objc_exception_throw ()
#2	0x00000001ac25ce1c in _CFThrowFormattedException ()
#3	0x00000001ac25a7d8 in -[__NSPlaceholderArray initWithObjects:count:].cold.3 ()
#4	0x00000001ac0d19c4 in -[__NSPlaceholderArray initWithObjects:count:] ()
#5	0x00000001ac1db1e0 in __createArray ()
#6	0x00000001ac0d8878 in +[NSArray arrayWithObject:] ()
#7	0x00000001ad5a8248 in +[NSOrthography defaultOrthographyForLanguage:] ()
#8	0x00000001014ac11c in +[NSOrthography(MGLAdditions) mgl_dominantScriptForMapboxStreetsLanguage:] at /Users/distiller/project/platform/ios/src/NSOrthography+MGLAdditions.m:8
[...]

PR #619, opened 2021-08-20
Avoid crash in NSOrthography category method

macdrevx:
Fixes https://github.com/mapbox/mapbox-gl-native-ios/issues/562

### DIFF
--- a/platform/ios/src/NSOrthography+MGLAdditions.m
+++ b/platform/ios/src/NSOrthography+MGLAdditions.m
@@ -2,7 +2,7 @@
 
 @implementation NSOrthography (MGLAdditions)
 
-+ (NSString *)mgl_dominantScriptForMapboxStreetsLanguage:(NSString *)language {
++ (nullable NSString *)mgl_dominantScriptForMapboxStreetsLanguage:(nullable NSString *)language {
     if (language == nil) {
         return nil;
     }
--- a/platform/ios/src/NSOrthography+MGLAdditions.m
+++ b/platform/ios/src/NSOrthography+MGLAdditions.m
@@ -3,6 +3,10 @@
 @implementation NSOrthography (MGLAdditions)
 
 + (NSString *)mgl_dominantScriptForMapboxStreetsLanguage:(NSString *)language {
+    if (language == nil) {
+        return nil;
+    }
+
     if (@available(iOS 11.0, *)) {
         NSLocale *locale = [NSLocale localeWithLocaleIdentifier:language];
         NSOrthography *orthography = [NSOrthography defaultOrthographyForLanguage:locale.localeIdentifier];
--- a/platform/darwin/src/NSString+MGLAdditions.h
+++ b/platform/darwin/src/NSString+MGLAdditions.h
@@ -22,14 +22,14 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  Returns a transliterated representation of the receiver using the specified
- script. If transliteration fails, the receiver will be returned.
+ script. If transliteration fails or script is nil, the receiver will be returned.
 
  Only supports scripts for languages used by Mapbox Streets.
 
  @param script The four-letter code representing the name of the script, as
     specified by ISO 15924.
  */
-- (NSString *)mgl_stringByTransliteratingIntoScript:(NSString *)script;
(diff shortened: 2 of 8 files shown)


nishant-karajgikar (approved):

platform/ios/src/NSOrthography+MGLAdditions.m
```diff
@@ -3,6 +3,10 @@
 @implementation NSOrthography (MGLAdditions)
 
 + (NSString *)mgl_dominantScriptForMapboxStreetsLanguage:(NSString *)language {
+    if (language == nil) {
```
1ec5, line 6: If `language` is `nil`, the reason is that `+[MGLVectorTileSource preferredMapboxStreetsLanguage]` returned `nil`. `-[MGLFeatureAccessibilityElement initWithAccessibilityContainer:feature:]` should use the `name` attribute instead of trying and failing to use the `name_(null)` attribute:

https://github.com/mapbox/mapbox-gl-native-ios/blob/ddf8e35970d1a8a46171bd798bcfa8f22d2760e0/platform/ios/src/MGLMapAccessibilityElement.mm#L51-L52

which would result in no accessibility label at all, assuming that `name` is present on the feature:

https://github.com/mapbox/mapbox-gl-native-ios/blob/ddf8e35970d1a8a46171bd798bcfa8f22d2760e0/platform/ios/src/MGLMapAccessibilityElement.mm#L70

platform/ios/src/NSOrthography+MGLAdditions.m
```diff
@@ -3,6 +3,10 @@
...
+        return nil;
```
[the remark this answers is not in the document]
macdrevx: Will update `-mgl_stringByTransliteratingIntoScript` to accept a nullable script string & will add corresponding assertion in the tests.

platform/ios/src/NSOrthography+MGLAdditions.m
macdrevx: When languageCode is null, nameAttribute ends up being `name_(null)`, and on line 53, name is nil. When name is nil, the fallback on line 56/57 is still applied.

I can still (and almost did initially) update this to be more defensive about `nil` `languageCode` though

macdrevx: Updated

platform/ios/src/NSOrthography+MGLAdditions.m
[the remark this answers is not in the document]
macdrevx: Updated

platform/ios/src/NSOrthography+MGLAdditions.m
1ec5: Yeah, I think that’s necessary because `name` is almost always present, so that fallback would rarely be applied.

platform/ios/src/NSOrthography+MGLAdditions.m
```diff
@@ -3,6 +3,10 @@
 @implementation NSOrthography (MGLAdditions)
 
 + (NSString *)mgl_dominantScriptForMapboxStreetsLanguage:(NSString *)language {
```
1ec5, line 5: By the way, my earlier feedback was indirectly suggesting that `NS_ASSUME_NONNULL_*` macros would be added to NSString+MGLAdditions.h and the parameter and return type of this method would be marked `nullable`.

1ec5 (approved):
Ideally we’d add nullability macros to NSString+MGLAdditions.h, but the changes here look good regardless.

platform/ios/src/NSOrthography+MGLAdditions.m
1ec5: Never mind, I misread this as a `== nil` rather than a `!= nil`:

https://github.com/mapbox/mapbox-gl-native-ios/blob/ddf8e35970d1a8a46171bd798bcfa8f22d2760e0/platform/ios/src/MGLMapAccessibilityElement.mm#L56

platform/ios/src/NSOrthography+MGLAdditions.m
macdrevx: Updated

PR #620, opened 2021-08-20
Invalidate unused offline packs before dealloc

macdrevx:
iOS 15 revealed a logic error in which certain offline packs were not being invalidated before being deallocated.

### DIFF
--- a/platform/darwin/src/MGLOfflinePack.mm
+++ b/platform/darwin/src/MGLOfflinePack.mm
@@ -147,11 +147,16 @@ private:
     MGLOfflineStorage *sharedOfflineStorage = [MGLOfflineStorage sharedOfflineStorage];
     __weak MGLOfflinePack *weakSelf = self;
     [sharedOfflineStorage getPacksWithCompletionHandler:^(NSArray<MGLOfflinePack *> *packs, __unused NSError * _Nullable error) {
+        MGLOfflinePack *strongSelf = weakSelf;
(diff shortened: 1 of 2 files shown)


platform/darwin/src/MGLOfflinePack.mm
```diff
@@ -147,11 +147,16 @@ - (void)reloadWithCompletionHandler:(void (^)(NSError * _Nullable error))complet
...
+        MGLOfflinePack *strongSelf = weakSelf;
         for (MGLOfflinePack *pack in packs) {
             if (pack.mbglOfflineRegion->getID() == regionID) {
-                weakSelf.mbglOfflineRegion = pack.mbglOfflineRegion;
-                break;
+                if (strongSelf.mbglOfflineRegion) {
+                    strongSelf->_mbglDatabaseFileSource->setOfflineRegionObserver(*strongSelf.mbglOfflineRegion, nullptr);
+                }
+                strongSelf.mbglOfflineRegion = pack.mbglOfflineRegion;
+                strongSelf->_mbglDatabaseFileSource->setOfflineRegionObserver(*strongSelf.mbglOfflineRegion, std::make_unique<MBGLOfflineDownloadObserver>(strongSelf));
```
macdrevx, line 157: Since we're swapping out the underlying offline region, it seems like we should also be removing and re-adding the observer.

1ec5 (approved):

nishant-karajgikar (approved):

